PR #89, opened 2017-09-11
Char array folding

### LINKED ISSUE #76: Char array folding
Parsing result of

```
char[] var = { 's', 't', 'r', 'i', 'g' };
```

should be

```JSON
{
  "NodeType": "AssignmentExpression",
  "Left": {
    "NodeType": "IdToken",
    "Id": "var"
  },
  "Right": {
    "NodeType": "StringLiteral",
    "Text": "strig"
  }
}
```

IliyVas:
fixes #76

### DIFF
--- a/Sources/PT.PM.Common/Helper.cs
+++ b/Sources/PT.PM.Common/Helper.cs
@@ -57,7 +57,7 @@ namespace PT.PM.Common
 
         static public bool TryCheckIdTokenValue(Expression expr, string value)
         {
-            if (expr == null || expr.NodeType != NodeType.IdToken)
+            if (expr?.NodeType != NodeType.IdToken)
             {
                 return false;
             }
--- a/Sources/PT.PM.Common/Helper.cs
+++ b/Sources/PT.PM.Common/Helper.cs
@@ -3,6 +3,8 @@ using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using PT.PM.Common.Nodes.Expressions;
+using PT.PM.Common.Nodes.Tokens;
 
 namespace PT.PM.Common
 {
@@ -31,7 +33,7 @@ namespace PT.PM.Common
 
         public static bool TryConvertToInt64(this string value, int fromBase, out long result)
         {
-            
+
             try
             {
                 result = Convert.ToInt64(value, fromBase);
@@ -51,6 +53,15 @@ namespace PT.PM.Common
                 int p = (int)Environment.OSVersion.Platform;
                 return (p == 4) || (p == 6) || (p == 128);
             }
+        }
+
+        static public bool TryCheckIdTokenValue(Expression expr, string value)
+        {
+            if (expr == null || expr.NodeType != NodeType.IdToken)
+            {
+                return false;
+            }
+            return ((IdToken)expr).TextValue == value;
         }
     }
 }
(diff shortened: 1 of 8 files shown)


Sources/PT.PM.Common/Helper.cs
```diff
@@ -52,5 +54,14 @@ public static bool IsRunningOnLinux
...
             }
         }
+
+        static public bool TryCheckIdTokenValue(Expression expr, string value)
+        {
+            if (expr?.NodeType != NodeType.IdToken)
+            {
+                return false;
+            }
+            return ((IdToken)expr).TextValue == value;
```
KvanTTT: Maybe replace with `return expr is IdToken idToken && idToken.TextValue == value`?

Sources/PT.PM.JavaParseTreeUst/Converter/StatementVisitor.cs
```diff
@@ -62,6 +63,23 @@ public UstNode VisitLocalVariableDeclaration(JavaParser.LocalVariableDeclaration
                 .Select(varDec => (AssignmentExpression)Visit(varDec))
                 .Where(initializer => initializer != null).ToArray();
 
+            if (initializers.Count() == 1 && initializers.First().Right is MultichildExpression multichildExpression)
```
KvanTTT: Simplify: 
```CSharp
if (initializers.SingleOrDefault()?.Right is MultichildExpression multichildExpression)
```

Sources/PT.PM.Common/Nodes/Expressions/MultichildExpression.cs
```diff
@@ -38,6 +38,13 @@ public MultichildExpression()
         {
         }
 
+        public int GetDepth(int multichildExpressionArgNumber)
+        {
+            return 1 +
+                (Expressions.Count() > 2 && Expressions[multichildExpressionArgNumber] is MultichildExpression child ?
+                    child.GetDepth(multichildExpressionArgNumber) : 0);
+        }
```
KvanTTT: * What the meaning of this code? It's unclear. Can you please add a comment for explanation.
* `multichildExpressionArgNumber` is an external parameter. Maybe use for safe `Expressions.ElementAtOrDefault(multichildExpressionArgNumber)`?

Sources/PT.PM.JavaParseTreeUst/Converter/StatementVisitor.cs
```diff
@@ -62,6 +63,23 @@ public UstNode VisitLocalVariableDeclaration(JavaParser.LocalVariableDeclaration
...
+            {
+                var expressions = multichildExpression.Expressions;
+                // is array?
+                if (Helper.TryCheckIdTokenValue(expressions.First(), "{") &&
+                    Helper.TryCheckIdTokenValue(expressions.Last(), "}"))
```
KvanTTT: Do `expressions` always have at least one element? Add an additional check otherwise or use `FirstOrDefault` and `LastOrDefault` methods.

Sources/PT.PM.JavaParseTreeUst/Converter/StatementVisitor.cs
IliyVas: SingeOrDefault throws an exception if there is more than one element in the sequence. It's unwanted behavior

KvanTTT: Ok.

PR #90, opened 2017-09-15
PatternVarOrFieldDeclaration

IliyVas:


Sources/PT.PM.Patterns/Nodes/PatternVarOrFieldDeclaration.cs
```diff
@@ -0,0 +1,151 @@
...
+using PT.PM.Common.Nodes.TypeMembers;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace PT.PM.Patterns.Nodes
+{
+    public class PatternVarOrFieldDeclaration : Expression
```
KvanTTT: Does `PatternVarOrFieldDeclaration` should be inherited from `Expression` i.e. entity that returns some result?
How difficult to fix it? I mean ordinary VarDeclaration too.

PR #93, opened 2017-10-05
PatternDeclarationUtils and PatternVarOrFieldDeclaration small fixes

IliyVas:


Sources/PT.PM.Matching/Patterns/PatternVarOrFieldDeclaration.cs
```diff
@@ -98,7 +98,15 @@ private MatchingContext EnumerateVarsOrFiels(List<AssignmentExpression> variable
...
-                MatchingContext match = Assignment.Match(variable, altContext);
+                MatchingContext match;
+                if (Assignment.Right != null)
+                {
+                    match = Assignment.Match(variable, altContext);
+                }
+                else
+                {
+                    match = Assignment.Left.Match(variable.Left, altContext);
+                }
```
KvanTTT: What purpose of this replacement? `PatternAssignmentExpression` have null checks: https://github.com/PositiveTechnologies/PT.PM/blob/dev/Sources/PT.PM.Matching/Patterns/PatternAssignmentExpression.cs#L39-L47

IliyVas: Without it `field <[ [Ll]og ]> <[ .+ ]>` pattern doesn't match `Logger log = Logger.getLogger(LoggingJul.class.getName());` because assigment Right expressions differs from each other

PR #97, opened 2017-10-09
Add PatternReturn, PatternThrow. Add posibility to match single expression in a method body

IliyVas:


Sources/PT.PM.Dsl.Tests/Data/ImproperValidationEmptyMethod.ptpm
```diff
@@ -1,3 +0,0 @@
-class : <[X509TrustManager|SSLSocketFactory]> {
-	<{ <[.+]>() {} }>
```
KvanTTT: I think we should use at least `<[\w+]>` syntax for identifiers especially for Java. This improves readability of identifiers and strings.

Sources/PT.PM.Dsl.Tests/Data/ImproperValidationEmptyMethod_full.ptpm
```diff
@@ -0,0 +1,3 @@
+class : <[X509TrustManager|SSLSocketFactory]> {
+	<{ (<[.+]>() {}) <|> (<[.+]>() { return null }) <|> (<[.+]>() { <{ throw new UnsupportedOperationException(<["\w+"]>) }> }) }>
```
KvanTTT: Сontrariwise, for strings it's more proper to use `<[".*"]>` syntax (sequence of arbitrary chars).

Sources/PT.PM.Dsl.Tests/DslParseTests.cs
```diff
@@ -41,7 +41,7 @@ public void Init()
         [TestCase("KeyManagementNullEncryptionKey.ptpm")]
         [TestCase("AttributesCodeInsideElementEvent.ptpm")]
         [TestCase("ExtendingSecurityManagerWithoutFinal.ptpm")]
-        [TestCase("ImproperValidationEmptyMethod.ptpm")]
+        [TestCase("ImproperValidationEmptyMethod_full.ptpm")]
```
KvanTTT: Common style: CommonCase for names. Remove underscore here.

Sources/PT.PM.Dsl/DslUstConverter.cs
```diff
@@ -175,10 +175,10 @@ public PatternUst VisitMethodDeclaration(DslParser.MethodDeclarationContext cont
             PatternUst result;
             IEnumerable<PatternUst> modifiers = ProcessLiteralsOrPatternIds(context._modifiers);
             PatternUst name = VisitLiteralOrPatternId(context.methodName);
-            var arbitraryDepthExpression = context.arbitraryDepthExpression();
-            if (arbitraryDepthExpression != null)
+            var bodyExpression = context.expression();
+            if (bodyExpression != null)
             {
-                var body = (PatternArbitraryDepth)VisitArbitraryDepthExpression(arbitraryDepthExpression);
+                var body = VisitExpression(bodyExpression);
```
KvanTTT: Use concrete type instead of `var`. The type is not obvious from the right part (method invocation).

Sources/PT.PM.Matching.Tests/JavaMatchingTests.cs
```diff
@@ -40,6 +40,12 @@ public void Match_TestPatternsJava_MatchedAllDefault()
                 Assert.Greater(matchingResults.Count(p => p.PatternKey == dto.Key), 0, dto.Description);
             }
 
+            var ImproperValidationEmptyMethod_partial = patternDtos.Single(dto => dto.Description.StartsWith("ImproperValidationEmptyMethod_partial"));
```
KvanTTT: Local vars starts from lower-case char and does not have underscores (common code style).

Sources/PT.PM.Matching/Patterns/PatternMethodDeclaration.cs
```diff
@@ -74,7 +74,23 @@ public override MatchingContext Match(Ust ust, MatchingContext context)
...
                     if (Body != null)
                     {
-                        newContext = Body.Match(methodDeclaration.Body, newContext);
+                        if (Body is PatternArbitraryDepth || Body is PatternStatements)
+                        {
+                            newContext = Body.Match(methodDeclaration.Body, newContext);
+                        }
+                        else
+                        {
+                            var statements = methodDeclaration.Body.Statements;
```
KvanTTT: Use more clear name `otherStatements` or something like this for getting rid of confusion between Pattern UST statements and UST statements.

Sources/PT.PM.Matching/Patterns/PatternReturn.cs
```diff
@@ -0,0 +1,47 @@
...
+                MatchingContext newContext = Pattern.Match(returnStatement.Return, context);
+                if (newContext.Success)
+                {
+                    return newContext.AddMatch(ust);
+                }
+                else
+                {
+                    return newContext.Fail();
+                }
+            }
```
KvanTTT: Use the following common style (see also other pattern nodes):

```
if (ust is ReturnStatement returnStatement)
{
    ...
}
else
{
    return newContext.Fail();
}
```

instead of early null check and break.

Sources/PT.PM.Matching/Patterns/PatternStatements.cs
```diff
@@ -46,7 +46,12 @@ public override MatchingContext Match(Ust ust, MatchingContext context)
 
             if (Statements == null || Statements.Count == 0)
             {
-                return context;
+                if(blockStatement.Statements == null || blockStatement.Statements.Count == 0)
+                {
+                    context = context.AddMatch(blockStatement);
+                    return context;
```
KvanTTT: Do not use early return for success matching. Success matches should be returned only in a single exit point (see other samples). Failed matches can be returned anywhere.

Sources/PT.PM.Matching/Patterns/PatternThrow.cs
```diff
@@ -0,0 +1,47 @@
...
+            : base(textSpan)
+        {
+            Pattern = pattern ?? throw new ArgumentNullException(nameof(pattern));
+        }
+
+        public override string ToString() => $"throw {Pattern}";
+
+        public override MatchingContext Match(Ust ust, MatchingContext context)
+        {
+            if (ust == null)
```
KvanTTT: Fix early null check.

Sources/PT.PM.Patterns/JavaPatterns.cs
```diff
@@ -448,8 +448,66 @@ public IEnumerable<PatternRoot> CreateJavaPatterns()
                     },
                     Body = new PatternArbitraryDepth
                     {
-                        Pattern = new PatternMethodDeclaration(
-                            Enumerable.Empty<PatternUst>().ToList(), new PatternIdRegexToken(".+"), false)
```
KvanTTT: In future, all UST and pattern nodes should take IEnumerable as an input parameter in constructors instead of other collections. So, `ToList()` invocation is redundant here.

Sources/PT.PM.Patterns/JavaPatterns.cs
```diff
@@ -448,8 +448,66 @@ public IEnumerable<PatternRoot> CreateJavaPatterns()
...
+                        Pattern = new PatternOr
+                        {
+                            Patterns = new List<PatternUst>
+                            {
+                                new PatternMethodDeclaration(Enumerable.Empty<PatternUst>().ToList(), new PatternIdRegexToken(".+"), false),
```
KvanTTT: Redundant `ToList()`.

Sources/PT.PM.Patterns/JavaPatterns.cs
```diff
@@ -448,8 +448,66 @@ public IEnumerable<PatternRoot> CreateJavaPatterns()
...
+                                            Name = new PatternIdRegexToken(".+"),
+                                            Body = new PatternReturn(new PatternNullLiteral())
+                                        },
+                                        new PatternMethodDeclaration
+                                        {
+                                            Modifiers = Enumerable.Empty<PatternUst>().ToList(),
+                                            Name = new PatternIdRegexToken(".+"),
+                                            Body = new PatternArbitraryDepth
+                                            {
+                                                Pattern = new PatternThrow(
```
KvanTTT: Move parenthesis to the next line (by analogy with curly brackets).

Sources/PT.PM.Dsl/DslUstConverter.cs
IliyVas: There can be any expression

KvanTTT: `PatternUst body = VisitExpression(bodyExpression);`

PR #99, opened 2017-10-09
Add PatternReturn, PatternThrow. Add posibility to match single expression in a method body

IliyVas:


Sources/PT.PM.Dsl.Tests/Data/ImproperValidationEmptyMethodFull.ptpm
```diff
@@ -0,0 +1,3 @@
+class : <[X509TrustManager|SSLSocketFactory]> {
+	<{ (<[\w+]>() {}) <|> (<[\w+]>() { return null }) <|> (<[\w+]>() { <{ throw new UnsupportedOperationException(<[".*"]>) }> }) }>
+}
```
KvanTTT: Maybe add line breaks for better readability? Something like this:

```DSL
class : <[X509TrustManager|SSLSocketFactory]> { <{
	(<[\w+]>() {}) <|>
	(<[\w+]>() { return null }) <|>
	(<[\w+]>() { <{ throw new UnsupportedOperationException(<[".*"]>) }> })
}> }
```

Sources/PT.PM.Matching/Patterns/PatternReturn.cs
```diff
@@ -0,0 +1,48 @@
+﻿using PT.PM.Common;
+using PT.PM.Common.Nodes;
+using PT.PM.Common.Nodes.Statements;
+using System;
+
+namespace PT.PM.Matching.Patterns
+{
+    public class PatternReturn : PatternUst
+    {
+        public PatternUst Pattern { get; set; } = PatternAny.Instance;
```
KvanTTT: As I understand `ReturnExpression` or just `Expression` is a more appropriate name instead of `Pattern`.

`Pattern` name used for artificial logical nodes (NOT).

Sources/PT.PM.Patterns/JavaPatterns.cs
```diff
@@ -448,8 +448,67 @@ public IEnumerable<PatternRoot> CreateJavaPatterns()
...
+                            Patterns = new List<PatternUst>
+                            {
+                                new PatternMethodDeclaration(Enumerable.Empty<PatternUst>(), new PatternIdRegexToken(@"\w+"), false),
+                                new PatternOr
+                                {
+                                    Patterns = new List<PatternUst>
+                                    {
+                                        new PatternMethodDeclaration
+                                        {
+                                            Modifiers = Enumerable.Empty<PatternUst>().ToList(),
```
KvanTTT: Why not a simple `new List<PatternUst>()`?
Actual for further cases too.

Sources/PT.PM.Matching/Patterns/PatternThrow.cs
```diff
@@ -0,0 +1,48 @@
+﻿using PT.PM.Common;
+using PT.PM.Common.Nodes;
+using PT.PM.Common.Nodes.Statements;
+using System;
+
+namespace PT.PM.Matching.Patterns
+{
+    public class PatternThrow : PatternUst
```
KvanTTT: Should be `PatternThrowStatement`.

Sources/PT.PM.Matching/Patterns/PatternThrow.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        public PatternUst Pattern { get; set; } = PatternAny.Instance;
```
KvanTTT: Should be `ThrowExpression` or `Expression`.

Sources/PT.PM.Matching/Patterns/PatternReturn.cs
```diff
@@ -0,0 +1,48 @@
+﻿using PT.PM.Common;
+using PT.PM.Common.Nodes;
+using PT.PM.Common.Nodes.Statements;
+using System;
+
+namespace PT.PM.Matching.Patterns
+{
+    public class PatternReturn : PatternUst
```
KvanTTT: `PatternReturnStatement` more clear.

Sources/PT.PM.Patterns/JavaPatterns.cs
IliyVas: I think it's slightly more efficient than `new List<PatternUst>()`.

KvanTTT: Why? `ToList()` creates a new instance anyway. Moreover, in this case, readability is preferable over performance.

IliyVas: Ok

PR #101, opened 2017-10-19
Fix single statement matching

IliyVas:


Sources/PT.PM.Dsl/DslUstConverter.cs
```diff
@@ -53,13 +53,20 @@ public PatternUst VisitDslCode(DslParser.DslCodeContext context)
             PatternUst result;
             if (context.statement().Length > 0)
             {
-                IEnumerable<PatternUst> statements = context.statement().Select(
+                List<PatternUst> statements = context.statement().Select(
```
KvanTTT: Why did you replace it on `List`? It's possible to operate `IEnumerable`.

PR #118, opened 2017-12-27
Determine empty string and reduced string textspans

IliyVas:


Sources/PT.PM.Matching/Patterns/PatternStringRegexLiteral.cs
```diff
@@ -31,8 +31,65 @@ public PatternStringRegexLiteral(Regex regex, TextSpan textSpan = default(TextSp
...
+            if (stringLiteral.InitialTextSpans.Any())
+            {
+                List<TextSpan> result = new List<TextSpan>();
+                stringLiteral.InitialTextSpans.Sort();
+
+                foreach (var location in matches)
+                {
+                    int offset = 0;
+                    int leftBound = 1;
+                    int rightBound = stringLiteral.InitialTextSpans[0].Length - 1; // - 2 because of quotes and then + 1
```
KvanTTT: Is it possible to use `EscapeCharsLength` property here?

Sources/PT.PM/UstSimplifier.cs
```diff
@@ -67,12 +67,17 @@ public override Ust Visit(BinaryOperatorExpression binaryOperatorExpression)
...
+                    initialTextSpans.Populate(rightString);
+
                     string resultText = leftString.Text + rightString.Text;
                     result = new StringLiteral
                     {
                         Text = resultText,
                         Root = binaryOperatorExpression.Root,
-                        TextSpan = leftExpression.TextSpan.Union(rightExpression.TextSpan)
+                        TextSpan = leftExpression.TextSpan.Union(rightExpression.TextSpan),
+                        InitialTextSpans = initialTextSpans
```
KvanTTT: We should copy collection here with respect to other methods.

Sources/PT.PM.Matching/Patterns/PatternStringRegexLiteral.cs
```diff
@@ -36,13 +36,15 @@ public override MatchContext Match(StringLiteral stringLiteral, MatchContext con
             if (stringLiteral.InitialTextSpans.Any())
             {
                 List<TextSpan> result = new List<TextSpan>();
-                stringLiteral.InitialTextSpans.Sort();
+                var initialTextSpans = stringLiteral.InitialTextSpans.OrderBy(el => el.Start).ToList();
```
KvanTTT: `TextSpan` implement `IComparable` interface. Why do not to use just `el => el`?

PR #129, opened 2018-04-08
Textspan fix

flippaa:
Implemented text loading for source code files enumerated in multiTextSpan nodes.

### DIFF
--- a/Sources/PT.PM/WorkflowResultBase.cs
+++ b/Sources/PT.PM/WorkflowResultBase.cs
@@ -12,7 +12,7 @@ namespace PT.PM
         where TStage : struct, IConvertible
         where TMatchResult : MatchResultBase<TPattern>
     {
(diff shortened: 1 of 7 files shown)


Sources/PT.PM/WorkflowResultBase.cs
```diff
@@ -222,7 +222,7 @@ protected void ThrowInvalidStageException(string stage)
             throw new InvalidOperationException($"Set {stage} as a final Stage or activate {nameof(IsIncludeIntermediateResult)} property");
         }
 
-        protected void AddEntity<T>(List<T> collection, T entity)
+        protected void AddEntity<T>(ICollection<T> collection, T entity)
```
KvanTTT: Use `IEnumerable<T>` instead of `ICollection<T>`.

KvanTTT:
Compilation error still exists. Please run and compile also PT.PM.Gui.sln.

PR #151, opened 2018-06-17
Fix #146 Simplify line-column text span string represantation

### LINKED ISSUE #146: Simplify line-column text span string represantation
* `[l,c)` - if start position equals to the end position.
* `[l,c1..c2)` - if start line equals to the end line, but columns are different.
* Maybe `[l1..l2,c)` if start line differs from the end, but columns are the same. But it's rare case.
* `[l1,c1..l2,c2)` - common format.

codeFather2:


Sources/PT.PM.Common/LineColumnTextSpan.cs
```diff
@@ -38,17 +38,23 @@ public LineColumnTextSpan(int beginLine, int beginColumn, int endLine, int endCo
...
+                    result = $"{result}{EndLine},";
+                }
+                if (BeginColumn != EndColumn)
+                {
+                    result = $"{result}{EndColumn}";
+                }
             }
 
+            result = $"{result})";
+
```
KvanTTT: I think the following code fragment is much clear:

```csharp
string result;
if (BeginLine == EndLine)
{
    result = BeginColumn == EndColumn
        ? $"[{BeginLine},{BeginColumn})"
        : $"[{BeginLine},{BeginColumn}..{EndColumn})";
}
else
{
    result = $"[{BeginLine},{BeginColumn}..{EndLine}{EndColumn})";
}
```

Sources/PT.PM.Tests/TextSpanTests.cs
```diff
@@ -19,15 +19,15 @@ public void Parse_TextSpan()
         }
 
         [Test]
```
KvanTTT: Excess `[Test]` attribute.

Sources/PT.PM.Common/LineColumnTextSpan.cs
```diff
@@ -38,15 +38,23 @@ public LineColumnTextSpan(int beginLine, int beginColumn, int endLine, int endCo
...
+                result = $"{result}..{EndLine},{EndColumn})";
+            }
+            else if (BeginColumn != EndColumn && BeginLine != EndLine)
+            {
+                result = $"{result},{BeginColumn}..{EndLine},{EndColumn})";
             }
             else
             {
-                result = $"[{BeginLine},{BeginColumn}..{EndLine},{EndColumn})";
+                result = $"{result},{BeginColumn})";
```
KvanTTT: I still prefer the following code:

```csharp
string result;

if (BeginLine == EndLine)
{
    result = BeginColumn == EndColumn
        ? $"[{BeginLine},{BeginColumn})"
        : $"[{BeginLine},{BeginColumn}..{EndColumn})";
}
else
{
    result = BeginColumn == EndColumn
        ? $"[{BeginLine}..{EndLine},{BeginColumn})"
        : $"[{BeginLine},{BeginColumn}..{EndLine},{EndColumn})";
}
```

Sources/PT.PM.Tests/TextSpanTests.cs
KvanTTT: Excess `[Test]` attribute.

PR #152, opened 2018-06-18
Make arguments more abstract

codeFather2:


Sources/PT.PM.Matching/Patterns/PatternInvocationExpression.cs
```diff
@@ -41,7 +49,15 @@ public override MatchContext Match(InvocationExpression invocation, MatchContext
...
             }
-            newContext = Arguments.Match(invocation.Arguments, newContext);
+
+            if (Arguments is PatternArgs patternArgs)
+            {
+                newContext = patternArgs.Match(invocation.Arguments, newContext);
+            }
+            else
+            {
+                newContext = context.Fail();
```
KvanTTT: I suggest change this string to `ewContext = Arguments.MatchUst(invocation.Arguments, newContext);`. It will  work for or logical operation.

Sources/PT.PM.Tests/UstPreprocessorTests.cs
```diff
@@ -103,7 +103,7 @@ public void Preprocess_MultiMultiPattern_RemovedDuplicates()
 
             var statements = ((PatternStatements)result).Statements;
             var invocation = (PatternInvocationExpression)statements.ElementAt(0);
-            Assert.AreEqual(1, invocation.Arguments.Args.Count(child => child is PatternMultipleExpressions));
+            Assert.AreEqual(1, (invocation.Arguments as PatternArgs)?.Args.Count(child => child is PatternMultipleExpressions));
```
KvanTTT: The type of `invocation.Arguments` is clearly defined. I suggest changing to `((PatternArgs)invocation.Arguments).Args`

Sources/PT.PM.Matching/Patterns/PatternInvocationExpression.cs
codeFather2: add this in last commit

Sources/PT.PM.Tests/UstPreprocessorTests.cs
codeFather2: Changed

PR #161, opened 2018-07-23
PatternAny by target source code and ust string

### LINKED ISSUE #160: Implement language-specific UST dump
If we convert the following PHP code

```php
$a->$b->$c
```

The following output from UST `.ToString()` we get:

```
a.b.c
```

It would be greate to use two modes: `Universal` and `Specific` for UST dump.

It can be used for more precise matching with `PatternAny` node.

Ideally it would be great to use [full-fidelity UST trees](https://github.com/dotnet/roslyn/wiki/Roslyn-Overview#syntax-trees).

KvanTTT:
* Correct RootUst text spans
* Match both by target source code and ust string, fixed #160 
* UstLinq: method names fixes (`Self` suffix)

### DIFF
--- a/Sources/PT.PM.CSharpParseTreeUst/Converter/CSharpRoslynParseTreeConverter.cs
+++ b/Sources/PT.PM.CSharpParseTreeUst/Converter/CSharpRoslynParseTreeConverter.cs
@@ -87,7 +87,10 @@ namespace PT.PM.CSharpParseTreeUst.RoslynUstVisitor
 
             if (root == null)
             {
-                root = new RootUst(children.FirstOrDefault()?.Root?.SourceCodeFile, CSharp.Language);
+                root = new RootUst(null, CSharp.Language)
+                {
+                    TextSpan = node.GetTextSpan()
+                };
             }
             root.Nodes = children.ToArray();
             return root;
--- a/Sources/PT.PM.Matching/PatternRoot.cs
+++ b/Sources/PT.PM.Matching/PatternRoot.cs
@@ -120,7 +120,7 @@ namespace PT.PM.Matching
                 }
 
                 var match = new MatchResult(
-                    ust is RootUst rootUst ? rootUst : ust.Root, context.PatternUst, context.Locations);
+                    ust.RootOrThis, context.PatternUst, context.Locations);
 
                 results.Add(match);
                 context.Logger.LogInfo(match);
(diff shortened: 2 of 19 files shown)


Sources/PT.PM.Matching/PatternRoot.cs
```diff
@@ -120,7 +120,7 @@ private static void TraverseChildren(PatternUst patternUst, Ust ust, MatchContex
                 }
 
                 var match = new MatchResult(
-                    ust is RootUst rootUst ? rootUst : ust.Root, context.PatternUst, context.Locations);
+                    ust.RootOrThis, context.PatternUst, context.Locations);
```
codeFather2: i think better way is use `ust.Root` property, and just override it in RootUst to `get => return this;`

Sources/PT.PM.CSharpParseTreeUst/Converter/CSharpRoslynParseTreeConverter.cs
```diff
@@ -87,7 +87,10 @@ public override Ust Visit(SyntaxNode node)
 
             if (root == null)
             {
-                root = new RootUst(children.FirstOrDefault()?.Root?.SourceCodeFile, CSharp.Language);
```
codeFather2: why you decide to remove `SourceCodeFile` from here?

Sources/PT.PM.Matching/PatternRoot.cs
KvanTTT: `RootUst` can have `Root` property itself in mixed-language files like JavaScript in PHP. In this case, JavaScript `RootUst` have a not null `Root` property that points to `RootUst` of PHP.

Sources/PT.PM.CSharpParseTreeUst/Converter/CSharpRoslynParseTreeConverter.cs
KvanTTT: I've discovered that it's `null` in any case. Thus there is no sense in such expression.

codeFather2: so `RootUst` can include private field `root`, and property can look like `get => root ?? this`

Sources/PT.PM.Matching/PatternRoot.cs
codeFather2: so RootUst can include private field root, and property can look like get => root ?? this.

PR #169, opened 2018-08-20
Long paths support (Windows)

### LINKED ISSUE #168: Support long paths (>= 260 chars) for net472 target runtime in Windows
See also [Why does the 260 character path length limit exist in Windows?](https://stackoverflow.com/q/1880321/1046374)

KvanTTT:
fixes #168

### DIFF
--- a/Sources/PT.PM.Common/CommonUtils.cs
+++ b/Sources/PT.PM.Common/CommonUtils.cs
@@ -3,14 +3,14 @@ using PT.PM.Common.Nodes.Expressions;
 using PT.PM.Common.Nodes.Tokens;
 using System;
 using System.Collections.Generic;
-using System.Threading;
+using System.Runtime.InteropServices;
 
-namespace PT.PM.Common
+namespace PT.PM.Common.Utils
 {
     public static class CommonUtils
     {
         private static bool? isRunningOnLinux = null;
-        private static bool? isSupportThreadAbort = null;
+        private static bool? isCoreApp = null;
 
         public const string Prefix = "pt.pm_";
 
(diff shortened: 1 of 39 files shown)


Sources/PT.PM.Common/CommonUtils.cs
```diff
@@ -43,39 +43,29 @@ public static bool TryConvertToInt64(this string value, int fromBase, out long r
...
+        public static bool IsWindows
         {
             get
             {
-                if (isRunningOnLinux == null)
+                if (!isRunningOnLinux.HasValue)
                 {
-                    int p = (int)Environment.OSVersion.Platform;
-                    isRunningOnLinux = (p == 4) || (p == 6) || (p == 128);
+                    isRunningOnLinux = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
```
codeFather2: i think it's bad idea to assign result of `IsOsPlatform(Windows)`, to `isRunningOnLinux`. Also we can have `OSX`. Better to call `IsOsPlatform(OsPlatform.Linux)`

Sources/PT.PM.Common/PathUtils.cs
```diff
@@ -0,0 +1,79 @@
...
+
+                return $@"\\?\{path}";
+            }
+
+            return path.NormalizeDirSeparator();
+        }
+
+        public static string NormalizeDirSeparator(this string path)
+        {
+            string notPlatformSeparator = CommonUtils.IsWindows ? "/" : "\\";
```
codeFather2: i think that `Path.AltDirectorySeparatorChar` property is better for this case, cause it contains '/' for Windows and Mac Os, and '\\' for Linux systems

Sources/PT.PM.Common/PathUtils.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            if (path.Contains(notPlatformSeparator))
```
codeFather2: you shouldn't check for separatorChar contains in path, `Replace` method will make it for you

Sources/PT.PM.Common/CommonUtils.cs
KvanTTT: Yes, it's a misprint.

Sources/PT.PM.Common/PathUtils.cs
KvanTTT: I hope this is an optimization, but not sure. It's rather rare case when slashes removed, so I decided to insert such a check. But maybe you are right and it's useless.

Sources/PT.PM.Common/PathUtils.cs
KvanTTT: Not correct: `AltDirectorySeparatorChar` is the same as `DirectorySeparatorChar` on Linux, see https://github.com/dotnet/docs/issues/3894

PR #172, opened 2018-09-24
Esprima JavaScript parser and converter

KvanTTT:
* Moved from ANTLR-base JavaScript parser to fast [Esprima](https://github.com/sebastienros/esprima-dotnet).
* Added new UST nodes
* Optimize memory consumption


Sources/PT.PM.Common/Nodes/Expressions/UnaryOperatorExpression.cs
```diff
@@ -36,7 +36,12 @@ public override string ToString()
             UnaryOperator op = Operator.UnaryOperator;
             if (UnaryOperatorLiteral.PrefixTextUnaryOperator.ContainsValue(op))
             {
-                return $"{Operator}{Expression}";
+                UnaryOperator unaryOp = Operator.UnaryOperator;
```
codeFather2: you already have this variable, it is declared above (i mean `UnaryOperator op` on line 36)

PR #174, opened 2018-10-05
PatternStatements.cs: dead code removal and simplification.

gabriel-fallen:
There was unused code in an `if` statement. I removed it and simplified the logic.

### DIFF
--- a/Sources/PT.PM.Matching/Patterns/PatternStatements.cs
+++ b/Sources/PT.PM.Matching/Patterns/PatternStatements.cs
@@ -47,7 +47,7 @@ namespace PT.PM.Matching.Patterns
 
             if (Statements == null || Statements.Count == 0)
             {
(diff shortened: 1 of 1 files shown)


Sources/PT.PM.Matching/Patterns/PatternStatements.cs
```diff
@@ -47,11 +47,7 @@ public override MatchContext Match(Ust ust, MatchContext context)
 
             if (Statements == null || Statements.Count == 0)
             {
-                if (blockStatement.Statements == null || blockStatement.Statements.Count == 0)
-                {
-                    newContext = newContext.AddMatch(blockStatement);
-                }
-                else
+                if (blockStatement.Statements != null && blockStatement.Statements.Count != 0)
```
KvanTTT: I think it can be simplified to `blockStatement?.Statements.Count > 0`

KvanTTT:
Welcome to PT.PM! Thank you for fix :)

PR #177, opened 2018-10-08
Tsql converter modification

codeFather2:


KvanTTT:
I suggest moving new Sql nodes from `Specific` directory to `Sql` directory.

Sources/PT.PM.Common/Nodes/Sql/QueryParameters.cs
```diff
@@ -0,0 +1,28 @@
+﻿using PT.PM.Common.Nodes.Collections;
+using PT.PM.Common.Nodes.Expressions;
+using System;
+using System.Collections.Generic;
+
+namespace PT.PM.Common.Nodes.Sql
+{
+    public class QueryParameters : Expression
```
KvanTTT: Why parameters inherited from `Expression`?

PR #180, opened 2018-10-19
Added condition which prevents exception for Substring method

flippaa:
That condition prevents exception for Substring method and saves matches


Sources/PT.PM.Common/CodeFile.cs
```diff
@@ -112,6 +112,8 @@ public string GetStringAtLine(int line)
 
         public string GetSubstring(TextSpan textSpan)
         {
+            if (textSpan.Start + textSpan.Length > Code.Length)
```
KvanTTT: `textSpan.End > Code.Length` would be a bit better :)

PR #184, opened 2018-10-27
Optimize ust simplification

KvanTTT:


Sources/PT.PM.Common/UstConstantFolder.cs
```diff
@@ -0,0 +1,281 @@
...
+            {
+                if (op.BinaryOperator == BinaryOperator.Plus || op.BinaryOperator == BinaryOperator.Concat)
+                {
+                    // Use StringBuilder instead of immutable strings and concatenation
+                    string rightString = rightValue.ToString();
+
+                    StringBuilder leftStringBuilder;
+                    if (leftValue is string leftString)
+                    {
+                        leftStringBuilder = new StringBuilder((leftString.Length + rightString.Length) * 2);
```
gabriel-fallen: I'm sorry I don't see why are you allocating twice as much space in a `StringBuilder` here?

Sources/PT.PM.Common/UstConstantFolder.cs
```diff
@@ -0,0 +1,281 @@
...
+                            case BinaryOperator.Divide:
+                                if (rightInt == 0)
+                                {
+                                    folded = false;
+                                    break;
+                                }
+                                resultInt = leftInt / rightInt;
+                                break;
+                            case BinaryOperator.Mod:
+                                resultInt = leftInt % rightInt;
```
gabriel-fallen: Don't we need to check `rightInt == 0` here too?

Sources/PT.PM.Common/UstConstantFolder.cs
```diff
@@ -0,0 +1,281 @@
...
+                    if (Logger.IsLogDebugs)
+                    {
+                        Logger.LogDebug(
+                            $"Strings {binaryOperatorExpression} concatenated to \"{leftStringBuilder}\" at {binaryOperatorExpression.TextSpan}");
+                    }
+                    
+                    return result;
+                }
+            }
+            else if (leftValue is long leftInt && rightValue is long rightInt)
```
gabriel-fallen: Will floating point values be hadled in future?

Sources/PT.PM.Common/UstConstantFolder.cs
KvanTTT: I hope this is a memory allocation optimization. Because we can use the rest `StringBuilder` capacity for further string foldings without reallocating. But it make sense for nested concatenations.

Consider the following expression: `"a" + "b" + "c" + "d"` and the sequence of foldings:

## Extra allocation

1. "a" (1) alloc
2. "b" (1) alloc
3. "ab" (4) alloc
4. "abc" (4) no alloc
5. "abcd" (4) no alloc

## No extra allocation

1. "a" (1) alloc
2. "b" (2) alloc
3. "ab" (2) alloc
4. "abc" (4) alloc
5. "abcd" (4) no alloc

Sources/PT.PM.Common/UstConstantFolder.cs
KvanTTT: Yep, you are right. BTW it is a compiler error:

image

Sources/PT.PM.Common/UstConstantFolder.cs
KvanTTT: I think yes but it's not so important for now.

Sources/PT.PM.Matching/PatternMatcher.cs
```diff
@@ -36,11 +38,13 @@ public List<MatchResult> Match(RootUst ust)
                     patterns = patterns.Where(pattern => pattern.FilenameWildcardRegex?.IsMatch(ust.SourceCodeFile.FullName) ?? true);
                 }
 
+                var ustFolder = IsFoldConstants ? new UstConstantFolder {Logger = Logger} : null;
```
gabriel-fallen: I don't get it, `ustFolder` here might be null when `IsFoldConstants` is `false`, but `MatchContext` throws `ArgumentNullException` when constant folder is `null`. Isn't it contradictory? At least it's confusing.

Sources/PT.PM.Common/UstConstantFolder.cs
gabriel-fallen: Yeah, but I don't see `leftStringBuilder` used for anything after you add `leftString` and `rightString` there. It goes into `ProcessBinaryExpression` but it doesn't modify the builder. Am I missing some other use?

gabriel-fallen: Ah, OK now I get it. I've failed to correctly consider recursive call. :smile:

Sources/PT.PM.Matching/PatternMatcher.cs
gabriel-fallen: It's reassuring to hear that! :smiley:

PR #191, opened 2018-12-06
Matching fixes

codeFather2:
Fixes for Matching of empty string

### DIFF
--- a/Sources/PT.PM.Common/Nodes/GeneralScope/TypeDeclaration.cs
+++ b/Sources/PT.PM.Common/Nodes/GeneralScope/TypeDeclaration.cs
@@ -12,23 +12,23 @@ namespace PT.PM.Common.Nodes.GeneralScope
 
         public List<TypeToken> BaseTypes { get; set; } = new List<TypeToken>();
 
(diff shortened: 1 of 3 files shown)


Sources/PT.PM.Common/Nodes/GeneralScope/TypeDeclaration.cs
```diff
@@ -12,23 +12,23 @@ public class TypeDeclaration : EntityDeclaration
 
         public List<TypeToken> BaseTypes { get; set; } = new List<TypeToken>();
 
-        public List<EntityDeclaration> TypeMembers { get; set; } = new List<EntityDeclaration>();
+        public List<Ust> TypeMembers { get; set; } = new List<Ust>();
```
gabriel-fallen: Why do you replace `EntityDeclaration` with `Ust` here?

codeFather2: In some cases we have `UsingDeclaration` in `TypeDeclaration`. `UsingDeclaration` is not `EntityDeclaration`

PR #195, opened 2018-12-13
ArrayCreationExpression KeywordNew NPE fix.

gabriel-fallen:
Silly me. Rather obvious mistake that somehow slipped tests...

### DIFF
--- a/Sources/PT.PM.Common/Nodes/Expressions/ArrayCreationExpression.cs
+++ b/Sources/PT.PM.Common/Nodes/Expressions/ArrayCreationExpression.cs
@@ -67,6 +67,7 @@ namespace PT.PM.Common.Nodes.Expressions
             var otherArrayCreation = (ArrayCreationExpression)other;
 
             int compareKeywordNewResult = (KeywordNew == null && otherArrayCreation.KeywordNew == null) ? 0 : KeywordNew?.CompareTo(otherArrayCreation.KeywordNew) ?? -1;
+
             if (compareKeywordNewResult != 0)
             {
                 return compareKeywordNewResult;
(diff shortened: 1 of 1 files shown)


Sources/PT.PM.Common/Nodes/Expressions/ArrayCreationExpression.cs
```diff
@@ -66,7 +66,7 @@ public override int CompareTo(Ust other)
 
             var otherArrayCreation = (ArrayCreationExpression)other;
 
-            int compareKeywordNewResult = KeywordNew.CompareTo(otherArrayCreation.KeywordNew);
+            int compareKeywordNewResult = (KeywordNew == null && otherArrayCreation.KeywordNew == null) ? 0 : KeywordNew?.CompareTo(otherArrayCreation.KeywordNew) ?? -1;
```
KvanTTT: Please add a line break for clearness

KvanTTT: Exactly! Thanks :)

Sources/PT.PM.Common/Nodes/Expressions/ArrayCreationExpression.cs
```diff
@@ -67,6 +67,7 @@ public override int CompareTo(Ust other)
             var otherArrayCreation = (ArrayCreationExpression)other;
 
             int compareKeywordNewResult = (KeywordNew == null && otherArrayCreation.KeywordNew == null) ? 0 : KeywordNew?.CompareTo(otherArrayCreation.KeywordNew) ?? -1;
+
```
KvanTTT: I mean the string with the code it too long. But I merge it - will fix it later.

PR #196, opened 2018-12-17
Setting KeywordNew on ArrayCreationExpression when convering C#.

gabriel-fallen:
`KeywordNew` distinguishes dynamically allocated arrays from statically allocated ones.


Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
```diff
@@ -123,6 +123,9 @@ public override Ust VisitArrayCreationExpression(ArrayCreationExpressionSyntax n
 
             var result = new ArrayCreationExpression(type, sizes, inits, node.GetTextSpan());
 
+            IdToken keywordNew = ConvertId(node.NewKeyword);
+            result.KeywordNew = keywordNew;
```
codeFather2: previous comment adressed to this code))

PR #199, opened 2018-12-26
UstConstantFolder fixes

codeFather2:
fix for array creation expression

### DIFF
--- a/Sources/PT.PM.Common/UstConstantFolder.cs
+++ b/Sources/PT.PM.Common/UstConstantFolder.cs
@@ -72,6 +72,7 @@ namespace PT.PM.Common
         private FoldResult TryFoldArrayCreationExpression(ArrayCreationExpression arrayCreationExpression)
         {
             if (arrayCreationExpression.Type?.TypeText == "char" &&
+                arrayCreationExpression.Initializers.Count > 0 &&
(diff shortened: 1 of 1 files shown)


Sources/PT.PM.Common/UstConstantFolder.cs
```diff
@@ -72,6 +72,7 @@ private FoldResult TryGetOrFold(Ust ust)
         private FoldResult TryFoldArrayCreationExpression(ArrayCreationExpression arrayCreationExpression)
         {
             if (arrayCreationExpression.Type?.TypeText == "char" &&
+                arrayCreationExpression.Initializers.Count > 0 &&
```
gabriel-fallen: What's the motivation I wander? Avoiding unnecessary empty builder allocation? Or it somehow breaks correctness?

codeFather2: before, with an empty collection, the method returns an `FoldResult` instance with an empty string inside, and this logic is wrong, cause if there is no initializers, method must return null value.

codeFather2: also, the type of the array element here can be not only a `char`, but also a `string`

PR #200, opened 2018-12-27
Matching even without this keyword as long as it's optional.

gabriel-fallen:
Special case inside `PatternMemberReferenceExpression.Match`.


gabriel-fallen:
At least it broke no *new* tests. :smile:

PR #202, opened 2019-01-11
Move type property below KeywordNew (automatic serialization bag fix)

codeFather2:


Sources/PT.PM.Common/Nodes/Expressions/ArrayCreationExpression.cs
```diff
@@ -6,10 +6,10 @@ namespace PT.PM.Common.Nodes.Expressions
 {
     public class ArrayCreationExpression : Expression
     {
-        public TypeToken Type { get; set; }
-
         public IdToken KeywordNew { get; set; }
 
+        public TypeToken Type { get; set; }
```
KvanTTT: How does this influence on serialization?

PR #203, opened 2019-01-17
Array initialization fixes for C#

codeFather2:


Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
```diff
@@ -488,10 +488,21 @@ public override Ust VisitObjectCreationExpression(ObjectCreationExpressionSyntax
...
+            var textSpan = node.GetTextSpan();
+
+            if (node.Kind() == SyntaxKind.ArrayInitializerExpression)
+            {
+                var sizes = new List<Expression> { new IntLiteral(children.Length) };
+                result = new ArrayCreationExpression(new TypeToken("", textSpan), sizes, children.ToList(), textSpan);
+            }
+            else
+            {
+                result = new MultichildExpression(children, node.GetTextSpan());
```
gabriel-fallen: Maybe `result = new MultichildExpression(children, textSpan);` for brevity?

Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
```diff
@@ -488,10 +488,21 @@ public override Ust VisitObjectCreationExpression(ObjectCreationExpressionSyntax
...
+            Ust result = null;
             var children = node.Expressions.Select(e => (Expression)VisitAndReturnNullIfError(e))
                 .ToArray();
-            
-            var result = new MultichildExpression(children, node.GetTextSpan());
+            var textSpan = node.GetTextSpan();
+
+            if (node.Kind() == SyntaxKind.ArrayInitializerExpression)
+            {
+                var sizes = new List<Expression> { new IntLiteral(children.Length) };
```
gabriel-fallen: 2D, 3D and so on array initializers are impossible?

codeFather2: In this kind of initializer i decide to represent 2d as 1d of 1d arrays and etc. I do this in order not to collect information from child arrays

Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
```diff
@@ -488,10 +488,21 @@ public override Ust VisitObjectCreationExpression(ObjectCreationExpressionSyntax
...
+                result = new ArrayCreationExpression(new TypeToken("", textSpan), sizes, children.ToList(), textSpan);
+            }
+            else
+            {
```
KvanTTT: What other alternatives are available here? Object constructor?

codeFather2: Initializers not currently supported. Dictionary initializer for example

PR #204, opened 2019-01-18
Add tuple support for C#

codeFather2:
Also add `TupleCreateExpression` as UstNode and `PatternTupleCreateExpression`

### DIFF
--- a/Sources/PT.PM.CSharpParseTreeUst.Tests/Data/ValueTuple.cs
+++ b/Sources/PT.PM.CSharpParseTreeUst.Tests/Data/ValueTuple.cs
@@ -3,5 +3,7 @@
     static void Main()
     {
         var tuple = (Password: "hardcoded", 1, 1.5, "otherstring");
+        (string password, int x) = ("hardcoded", 1);
+        (string password, int x) tuple = ("hardcoded", 1);
     }
 }
--- a/Sources/PT.PM.CSharpParseTreeUst.Tests/Data/ValueTuple.cs
+++ b/Sources/PT.PM.CSharpParseTreeUst.Tests/Data/ValueTuple.cs
@@ -1,0 +1,7 @@
+﻿class TestTuples
+{
+    static void Main()
+    {
+        var tuple = (Password: "hardcoded", 1, 1.5, "otherstring");
+    }
+}
--- a/Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
+++ b/Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
@@ -91,6 +91,7 @@ namespace PT.PM.CSharpParseTreeUst.RoslynUstVisitor
                 };
 
                 var defaultName = new IdToken($"Item{i + 1}", assignment.TextSpan);
+                result.Initializers.Add(assignment);
 
                 if (arg.NameColon != null)
                 {
@@ -109,8 +110,6 @@ namespace PT.PM.CSharpParseTreeUst.RoslynUstVisitor
                 {
                     assignment.Left = defaultName;
                 }
-
-                result.Initializers.Add(assignment);
             }
 
             return result;
--- a/Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
+++ b/Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
@@ -90,19 +90,27 @@ namespace PT.PM.CSharpParseTreeUst.RoslynUstVisitor
                     TextSpan = arg.GetTextSpan()
                 };
 
-                assignment.Left = new IdToken($"Item{i + 1}", assignment.TextSpan);
-                result.Initializers.Add(assignment);
+                var defaultName = new IdToken($"Item{i + 1}", assignment.TextSpan);
 
                 if (arg.NameColon != null)
                 {
-                    var name = (Expression)VisitAndReturnNullIfError(arg.NameColon.Name);
+                    var name = (IdToken)VisitAndReturnNullIfError(arg.NameColon.Name);
+
+                    assignment.Left = name;
+
                     result.Initializers.Add(new AssignmentExpression
                     {
-                        Left = name,
-                        Right = new MemberReferenceExpression(null, new IdToken($"Item{i + 1}", name.TextSpan), name.TextSpan),
+                        Left = new IdToken(defaultName.Id, defaultName.TextSpan),
+                        Right = new MemberReferenceExpression(null, new IdToken(name.Id, name.TextSpan), name.TextSpan),
                         TextSpan = assignment.TextSpan
                     });
                 }
+                else
+                {
+                    assignment.Left = defaultName;
+                }
+
+                result.Initializers.Add(assignment);
             }
 
             return result;
@@ -381,7 +389,6 @@ namespace PT.PM.CSharpParseTreeUst.RoslynUstVisitor
         {
             var alias = (IdToken)VisitIdentifierName(node.Alias);
             var name = (Token)base.Visit(node.Name);
-
             var result = new TypeToken(new[] { alias.Id, name.TextValue }, node.GetTextSpan());
             return result;
         }
@@ -419,7 +426,6 @@ namespace PT.PM.CSharpParseTreeUst.RoslynUstVisitor
         {
             var target = (Expression)base.Visit(node.Expression);
             ArgsUst args = node.ArgumentList == null ? null : (ArgsUst)VisitBracketedArgumentList(node.ArgumentList);
-
             var result = new IndexerExpression(target, args, node.GetTextSpan());
             return result;
         }
--- a/Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
+++ b/Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
@@ -89,15 +89,20 @@ namespace PT.PM.CSharpParseTreeUst.RoslynUstVisitor
                     Right = (Expression)VisitAndReturnNullIfError(arg.Expression),
                     TextSpan = arg.GetTextSpan()
                 };
-                if (arg.NameColon == null)
-                {
-                    assignment.Left = new IdToken($"Item{i + 1}");
-                }
-                else
-                {
-                    assignment.Left = (Expression)VisitAndReturnNullIfError(arg.NameColon.Name);
-                }
+
+                assignment.Left = new IdToken($"Item{i + 1}", assignment.TextSpan);
                 result.Initializers.Add(assignment);
+
+                if (arg.NameColon != null)
+                {
+                    var name = (Expression)VisitAndReturnNullIfError(arg.NameColon.Name);
+                    result.Initializers.Add(new AssignmentExpression
+                    {
+                        Left = name,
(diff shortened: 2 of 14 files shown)


Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
```diff
@@ -73,6 +73,36 @@ public override Ust VisitAnonymousObjectCreationExpression(AnonymousObjectCreati
...
+            {
+                var arg = node.Arguments[i];
+                var assignment = new AssignmentExpression
+                {
+                    Right = (Expression)VisitAndReturnNullIfError(arg.Expression),
+                    TextSpan = arg.GetTextSpan()
+                };
+                if (arg.NameColon == null)
+                {
+                    assignment.Left = new IdToken($"Item{i + 1}");
```
gabriel-fallen: Strictly speaking `ItemN` name gets generated always. Won't it cause problems with cases like
```csharp
var data = (42, Payload: source());
pvf(data.Item2); // the same as pvf(data.Payload)
```
will it?

Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
```diff
@@ -328,8 +358,8 @@ public override Ust VisitAliasQualifiedName(AliasQualifiedNameSyntax node)
         {
             var alias = (IdToken)VisitIdentifierName(node.Alias);
             var name = (Token)base.Visit(node.Name);
-        
-            var result = new TypeToken(new [] { alias.Id, name.TextValue }, node.GetTextSpan());
+
```
gabriel-fallen: Why empty line at all?

Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
```diff
@@ -387,7 +417,7 @@ public override Ust VisitImplicitElementAccess(ImplicitElementAccessSyntax node)
 
         public override Ust VisitIndexerMemberCref(IndexerMemberCrefSyntax node)
         {
-            
+
```
gabriel-fallen: Does this empty line make sense?

Sources/PT.PM.Common/Nodes/Expressions/TupleCreateExpression.cs
```diff
@@ -0,0 +1,20 @@
...
+            if (Initializers != null)
+            {
+                return Initializers.ToArray();
+            }
+            return new Ust[0];
+        }
+
+        public override Expression[] GetArgs()
+        {
+            return new Expression[0];
```
gabriel-fallen: Maybe throwing an exception would be more appropriate?

Sources/PT.PM.CSharpParseTreeUst.Tests/Data/ValueTuple.cs
```diff
@@ -0,0 +1,7 @@
+﻿class TestTuples
+{
+    static void Main()
+    {
+        var tuple = (Password: "hardcoded", 1, 1.5, "otherstring");
```
KvanTTT: What about unit tests like this?

```csharp
(a, b) = ("a", "b");
```

Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
```diff
@@ -73,6 +73,41 @@ public override Ust VisitAnonymousObjectCreationExpression(AnonymousObjectCreati
...
+                assignment.Left = new IdToken($"Item{i + 1}", assignment.TextSpan);
+                result.Initializers.Add(assignment);
+
+                if (arg.NameColon != null)
+                {
+                    var name = (Expression)VisitAndReturnNullIfError(arg.NameColon.Name);
+                    result.Initializers.Add(new AssignmentExpression
+                    {
+                        Left = name,
+                        Right = new MemberReferenceExpression(null, new IdToken($"Item{i + 1}", name.TextSpan), name.TextSpan),
```
gabriel-fallen: Is it necessary to create new `IdToken`? Can't we reuse `assignment.Left`?

Besides, I don't see the point of `MemberReferenceExpression` with `null` target, can't we just reuse `assignment.Right` here too? Like this:
```
result.Initializers.Add(new AssignmentExpression
{
    Left = name,
    Right = assignment.Right,
    TextSpan = assignment.TextSpan
}
```

Sources/PT.PM.CSharpParseTreeUst/Converter/Helper.cs
```diff
@@ -35,6 +41,100 @@ protected TypeToken ConvertType(Ust node)
             return null;
         }
 
+        /// <summary>
+        /// Performs deconstruction of the tuples, in the absence of a name, else create vitual memrefs
```
gabriel-fallen: "When a name is absent, deconstruct a given tuple, otherwise create virtual memrefs."

What is a "memref" by the way? :)

Sources/PT.PM.CSharpParseTreeUst/Converter/Helper.cs
```diff
@@ -35,6 +41,100 @@ protected TypeToken ConvertType(Ust node)
...
+                    continue;
+                }
+
+                CreateVirtualMemRefs(ref variables, typeElement, i, textSpan);
+            }
+            var declaration = new VariableDeclarationExpression(null, variables, textSpan);
+            return new ExpressionStatement(declaration);
+        }
+
+        private void CreateVirtualMemRefs(ref List<AssignmentExpression> variables, TupleElementSyntax typeElement, int initializerNumber, TextSpan textSpan)
```
gabriel-fallen: I suggest renaming `MemRefs` to `MemberRefs` not to confuse with "memory ref".

Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
codeFather2: yes, it necessary) all tokens should be unique

Sources/PT.PM.CSharpParseTreeUst/Converter/Helper.cs
codeFather2: member reference)

Sources/PT.PM.CSharpParseTreeUst/Converter/Helper.cs
codeFather2: good idea, thanks

Sources/PT.PM.CSharpParseTreeUst/Converter/Helper.cs
```diff
@@ -35,6 +41,100 @@ protected TypeToken ConvertType(Ust node)
...
+                    continue;
+                }
+
+                CreateVirtualMemberRefs(ref variables, typeElement, i, textSpan);
+            }
+            var declaration = new VariableDeclarationExpression(null, variables, textSpan);
+            return new ExpressionStatement(declaration);
+        }
+
+        private void CreateVirtualMemberRefs(ref List<AssignmentExpression> variables, TupleElementSyntax typeElement, int initializerNumber, TextSpan textSpan)
```
KvanTTT: Make this method pure: remove `ref` and return a new `List<AssignmentExpression>` instead.

Sources/PT.PM.CSharpParseTreeUst/Converter/Helper.cs
```diff
@@ -35,6 +41,100 @@ protected TypeToken ConvertType(Ust node)
...
+        {
+            variables.ForEach(variable =>
```
KvanTTT: Do not use linq methods with side effects, specifically `ForEach`. Use an ordinary `foreach` loop instead.

Sources/PT.PM.CSharpParseTreeUst/Converter/Helper.cs
```diff
@@ -35,6 +41,100 @@ protected TypeToken ConvertType(Ust node)
...
+            {
+                var tupleTypeElementMemRef = new MemberReferenceExpression
+                {
+                    Target = new IdToken(((IdToken)variable.Left).Id, variable.Left.TextSpan),
+                    Name = ConvertId(typeElement.Identifier)
+                };
+
+                var right = (TupleCreateExpression)variable?.Right;
```
KvanTTT: Maybe it's safer using `variable?.Right as TupleCreateExpression` here and in similar cases?

Sources/PT.PM.Matching.Tests/CSharpMatchingTests.cs
```diff
@@ -59,5 +66,8 @@ public void Match_HardcodedPasswordAspx_WithoutException()
             match = matchResults.ElementAt(1).MatchedCode;
             Assert.IsTrue(match.Contains("PASSWORD") && match.Contains("hardcoded"));
         }
+
+        private FileCodeRepository CreateTestFileCodeRepo(string fileName)
```
KvanTTT: In other places, we don't shorten the repository word. Replace with `CreateTestFileCodeRepository`.

Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
gabriel-fallen: Then at least put that comment regarding `null` in the code. :wink: 

But I still think it's better to just use `assignment.Right` instead.

Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
```diff
@@ -73,6 +73,39 @@ public override Ust VisitAnonymousObjectCreationExpression(AnonymousObjectCreati
...
+                    TextSpan = arg.GetTextSpan()
+                };
+                result.Initializers.Add(assignment);
+
+                if (arg.NameColon != null)
+                {
+                    result.Initializers.Add(new AssignmentExpression
+                    {
+                        Left = (IdToken)VisitAndReturnNullIfError(arg.NameColon.Name),
+                        Right = (Expression)VisitAndReturnNullIfError(arg.Expression),
```
gabriel-fallen: Wouldn't something like `assignment.Right.clone()` be more efficient than rewalking the whole subtree?

Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
```diff
@@ -177,10 +210,28 @@ public override Ust VisitOmittedArraySizeExpression(OmittedArraySizeExpressionSy
 
         public override Ust VisitAssignmentExpression(AssignmentExpressionSyntax node)
         {
+            var textSpan = node.GetTextSpan();
+            if (node.Left is TupleExpressionSyntax tupleLeft
+                && node.Right is TupleExpressionSyntax tupleRight)
+            {
+                var assignments = new List<AssignmentExpression>();
```
KvanTTT: For the future: if the number of elements is defined, initialize the list with it: `var assignments = new List<AssignmentExpression>(tupleLeft.Arguments.Count);`

Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
codeFather2: there can be quite complex structures and copying may not give a satisfying result, in contrast to the rewalking

PR #205, opened 2019-01-28
MessagePack

### LINKED ISSUE #198: Analyze different binary serializers and choose the appropriate one
### Requirements

* C#, C++ support (ideally, Python, Go and other widespread and potentially supported languages)
* Zero allocation memory consumption
* Good performance
* No need for attributes in UST and other classes to serialize/deserialize

### Candidates

* [protobuf](https://github.com/protocolbuffers/protobuf) (Google)
* [flatbuffers](https://google.github.io/flatbuffers/) (Google)
* [bond](https://microsoft.github.io/bond/) (Miscrosoft)
* [BSON](http://bsonspec.org/). Binary Json. Advantage: our current library [Json.NET](https://www.newtonsoft.com/json) supports this format.
* [MessagedPack](https://msgpack.org/index.html). It's used by UI team
* Other libraries

### References

* [Comparison of data serialization formats](https://en.wikipedia.org/wiki/Comparison_of_data_serialization_formats) (Wikipedia)
* [Data Serialization Comparison: JSON, YAML, BSON, MessagePack](https://www.sitepoint.com/data-serialization-comparison-json-yaml-bson-messagepack/)

Also, at first maybe it makes sense to try fast and zero allocation JSON serialization library [Utf8Json](https://github.com/neuecc/Utf8Json).

KvanTTT:
* Introduce `IFile` interface and inherited classes `TextFile`, `BinaryFile`;
* Add `int Key` to `Ust`, generate unique identifiers for all ust;
* Language class replaced with enum;
* Implemented MessagePack serializators, fixes #198;
* Add `--serialize-format` cli option (`json`, `msgpack`).

### DIFF
--- a/Sources/PT.PM.Common/Nodes/Expressions/Expression.cs
+++ b/Sources/PT.PM.Common/Nodes/Expressions/Expression.cs
@@ -7,7 +7,7 @@ using PT.PM.Common.Nodes.Tokens.Literals;
 namespace PT.PM.Common.Nodes.Expressions
 {
     [MessagePackObject]
-    
+
     [Union((int)NodeType.AnonymousMethodExpression, typeof(AnonymousMethodExpression))]
     [Union((int)NodeType.AnonymousObjectExpression, typeof(AnonymousObjectExpression))]
     [Union((int)NodeType.ArgumentExpression, typeof(ArgumentExpression))]
@@ -21,11 +21,12 @@ namespace PT.PM.Common.Nodes.Expressions
     [Union((int)NodeType.MemberReferenceExpression, typeof(MemberReferenceExpression))]
     [Union((int)NodeType.MultichildExpression, typeof(MultichildExpression))]
     [Union((int)NodeType.ObjectCreateExpression, typeof(ObjectCreateExpression))]
+    [Union((int)NodeType.TupleCreateExpression, typeof(TupleCreateExpression))]
     [Union((int)NodeType.UnaryOperatorExpression, typeof(UnaryOperatorExpression))]
     [Union((int)NodeType.VariableDeclarationExpression, typeof(VariableDeclarationExpression))]
     [Union((int)NodeType.WrapperExpression, typeof(WrapperExpression))]
     [Union((int)NodeType.YieldExpression, typeof(YieldExpression))]
-    
+
     [Union((int)NodeType.BaseReferenceToken, typeof(BaseReferenceToken))]
     [Union((int)NodeType.IdToken, typeof(IdToken))]
     [Union((int)NodeType.ThisReferenceToken, typeof(ThisReferenceToken))]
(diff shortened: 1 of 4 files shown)


Sources/PT.PM.Common/MessagePack/RootUstFormatter.cs
```diff
@@ -0,0 +1,146 @@
...
+                {
+                    if (textSpan.File != null && !localSourceFiles.Contains(textSpan.File))
+                    {
+                        localSourceFiles.Add(textSpan.File);
+                    }
+                }
+            });
+
+            var textSpanFormatter = formatterResolver.GetFormatter<TextSpan>();
+            if (textSpanFormatter is TextSpanFormatter formatter)
```
codeFather2: Are there cases when in this place the formatter is not a `TextSpanFormatter`?

Sources/PT.PM.Common/Nodes/Tokens/Literals/Literal.cs
```diff
@@ -1,5 +1,17 @@
-﻿namespace PT.PM.Common.Nodes.Tokens.Literals
+﻿using MessagePack;
+
+namespace PT.PM.Common.Nodes.Tokens.Literals
 {
+    [MessagePackObject]
+    [Union((int)NodeType.BinaryOperatorLiteral, typeof(BinaryOperatorLiteral))]
```
codeFather2: Is it necessary to duplicate the `Union` attribute at all stages of class inheritance?

Sources/PT.PM.Common/PathUtils.cs
```diff
@@ -37,11 +36,14 @@ private static string NormalizePath(this string path, bool isDirectory = true, b
...
-            string notPlatformSeparator = CommonUtils.IsWindows ? "/" : "\\";
+            char notPlatformSeparator = CommonUtils.IsWindows ? '/' : '\\';
 
-            if (path.Contains(notPlatformSeparator))
+            foreach (char c in path)
             {
-                return path.Replace(notPlatformSeparator, Path.DirectorySeparatorChar.ToString());
+                if (c == notPlatformSeparator)
+                {
+                    return path.Replace(notPlatformSeparator, Path.DirectorySeparatorChar);
```
codeFather2: Can be replaced with `path.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar)`. this will help remove the foreach loop

Sources/PT.PM.Common/MessagePack/RootUstFormatter.cs
KvanTTT: No. But I like such a style, because it's safer in general.

Sources/PT.PM.Common/MessagePack/RootUstFormatter.cs
```diff
@@ -0,0 +1,146 @@
...
+
+                var filesFormatter = formatterResolver.GetFormatter<TextFile[]>();
+                TextFile[] localSourceFiles = filesFormatter.Deserialize(bytes, newOffset, formatterResolver, out size);
+                newOffset += size;
+
+                lock (SourceFiles)
+                {
+                    foreach (TextFile localSourceFile in localSourceFiles)
+                    {
+                        SourceFiles.Add(localSourceFile);
```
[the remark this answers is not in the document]
KvanTTT: Not possible, because HashSet does not have such a method.

Sources/PT.PM.Common/Nodes/Tokens/Literals/Literal.cs
KvanTTT: Unfortunately yes. Also, there is a special test `Check_MessagePackTypes` for checking if all abstract classes mapped to concrete ones.

Sources/PT.PM.Common/PathUtils.cs
KvanTTT: My motivation: do not do excess string allocations. But maybe you are right and it does not make sense.

PR #206, opened 2019-02-04
StartStage removed, GUI fixed

KvanTTT:
* Removed --start-stage: detect languages including JSON and MessagePack automatically
* Fixes GUI for working with JSON files

### DIFF
--- a/Sources/PT.PM.Common/LanguageUtils.cs
+++ b/Sources/PT.PM.Common/LanguageUtils.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using static PT.PM.Common.Language;
 
 namespace PT.PM.Common
 {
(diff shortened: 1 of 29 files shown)


codeFather2:
totally approve!!

Sources/PT.PM.Common/LanguageUtils.cs
```diff
@@ -102,7 +128,7 @@ public static Language[] GetLanguagesByExtension(string extension)
 
         public static bool HasAntlrParser(this Language language) => LanguageInfos[language].HasAntlrParser;
 
-        public static bool IsParserExists(this Language language) => LanguagesWithParser.Contains(language);
+        public static bool IsParserExists(this Language language) => LanguagesWithParser.Contains(language) || language.IsSerialization();
```
gabriel-fallen: Why not just add `SerializationLanguages` to `LanguagesWithParser`?

Sources/PT.PM.Matching/MatchResult.cs
```diff
@@ -27,14 +23,13 @@ public MatchResult(TextFile sourceFile, PatternRoot pattern, IEnumerable<TextSpa
...
-                Suppressed = PatternUtils.IsSuppressed(SourceFile,
-                    lastTextSpan.GetLineColumnTextSpan(SourceFile));
+                Suppressed = PatternUtils.IsSuppressed(SourceFile, lastTextSpan.GetLineColumnTextSpan(SourceFile));
             }
         }
 
         public override string ToString()
         {
-            return $"Pattern {Pattern} mathched at {(string.Join(", ", TextSpans.Select(textSpan => SourceFile.GetLineColumnTextSpan(textSpan))))}";
+            return $"Pattern {Pattern} matched at {(string.Join(", ", TextSpans.Select(textSpan => SourceFile.GetLineColumnTextSpan(textSpan))))}";
```
gabriel-fallen: I think just "match" would be even better. :)

Sources/PT.PM.Common/LanguageUtils.cs
KvanTTT: Because serialization languages (json, msgpack) are being processed in another way and their parsers are not being registered in [RegisterAllParsersAndCovnerters method](https://github.com/PositiveTechnologies/PT.PM/blob/dev/Sources/PT.PM/Utils.cs#L27). That's why `LanguagesWithParser` does not contain parsers for Json and MessagePack.

But I'll think whether I can refactor this moment.

KvanTTT: Finally, I decided to rename this method to `IsParserExistsOrSerialized`.

gabriel-fallen: `IsParserExistsOrSerialized` reads like "is parser exists or the parser is serialized" which isn't really accurate. :)
What's the reason they are combined? What's the common processing?

KvanTTT: You are right. Correct identifier: `IsParserExistsOrSerializing`. This means "is a parser for language exists or is language serializing`.

PR #208, opened 2019-02-15
Message pack improvements

KvanTTT:
* Improved MessagePack Ust serialization, unified and more compact. The fully manual serialization with help of reflection is used
* `Parent` property returned to `Ust` for convenience


Sources/PT.PM.Common/MessagePack/RootUstMessagePackSerializer.cs
```diff
@@ -69,52 +73,423 @@ public static byte[] Serialize(RootUst rootUst, bool isLineColumn, bool compress
...
+
+            var localSourceFiles = (TextFile[])DeserializeObject(bytes, newOffset, typeof(TextFile[]), out size);
+            newOffset += size;
+            textSpanSerializer.LocalSourceFiles = localSourceFiles;
+
+            lock (sourceFiles)
+            {
+                foreach (TextFile localSourceFile in localSourceFiles)
+                {
+                    sourceFiles.Add(localSourceFile);
```
gabriel-fallen: Are duplicates allowed in `sourceFiles`? Could there be duplicates if we unconditionally add files from different threads?

Sources/PT.PM.Common/MessagePack/RootUstMessagePackSerializer.cs
```diff
@@ -69,52 +73,423 @@ public static byte[] Serialize(RootUst rootUst, bool isLineColumn, bool compress
...
+            {
+                return MessagePackBinary.WriteSingle(ref bytes, offset, (float) value);
+            }
+
+            if (type == typeof(DateTime))
+            {
+                return MessagePackBinary.WriteDateTime(ref bytes, offset, (DateTime) value);
+            }
+
+            throw new NotImplementedException($"Serialization of {type.Name} is not supported");
```
gabriel-fallen: No need for `uint`s? :smiley:

Sources/PT.PM.Common/MessagePack/RootUstMessagePackSerializer.cs
```diff
@@ -1,22 +1,29 @@
...
 
-        private TextSpanFormatter textSpanFormatter;
-        private FileFormatter fileFormatter;
-        private RootUstFormatter rootUstFormatter;
+        private BinaryFile serializedFile;
+        private TextSpanSerializer textSpanSerializer;
+        private FileSerializer fileSerializer;
+
+        private readonly Stack<RootUst> rootAncestors = new Stack<RootUst>();
+        private readonly Stack<Ust> ancestors = new Stack<Ust>();
```
gabriel-fallen: It's unclear how `ancestors` and `rootAncestors` work, some comments would be very useful.

Sources/PT.PM.Common/MessagePack/TextSpanSerializer.cs
```diff
@@ -77,15 +75,15 @@ public int Serialize(ref byte[] bytes, int offset, TextSpan value, IFormatterRes
             if (fileIndex == -1)
             {
                 Logger.LogError(new ReadException(SerializedFile, message:
-                    $"{nameof(TextSpan.File)} of {nameof(TextSpan)} {value} is not correctly mapped in {nameof(RootUstFormatter)}"));
+                    $"{nameof(TextSpan.File)} of {nameof(TextSpan)} {value} is not correctly mapped"));
```
gabriel-fallen: "... is mapped incorrectly"?

Sources/PT.PM.Common/MessagePack/RootUstMessagePackSerializer.cs
flippaa: I suppose that `localSourceFiles` collection will be used for both **Dfg** and **Ust** `TextFile` mappings, won't it?

Sources/PT.PM.Common/MessagePack/RootUstMessagePackSerializer.cs
KvanTTT: Not now, but later it's possible. I did it just in case.

Sources/PT.PM.Common/Nodes/RootUst.cs
```diff
@@ -9,31 +9,29 @@
...
     {
         private Language[] sublanguges;
 
         [Key(UstFieldOffset)]
-        public Language Language { get; }
+        public Language Language { get; set; }
 
-        [JsonProperty("SourceCodeFile")] // TODO: back compatibility with external serializers
-        [Key(UstFieldOffset + 1)]
+        [JsonProperty("SourceCodeFile"), JsonIgnore, IgnoreMember] // TODO: back compatibility with external serializers
```
gabriel-fallen: What's the point in keeping `JsonProperty` attribute with `JsonIgnore`?

Sources/PT.PM.Common/Reflection/ReflectionCache.cs
```diff
@@ -1,53 +1,173 @@
...
-                    if (assembly.IsActual())
+                    if (assemblyType.Name.Equals(typeStr))
                     {
-                        foreach (Type type in assembly.GetTypes()
-                            .Where(myType => myType.IsClass && !myType.IsAbstract &&
-                            myType.GetInterfaces().Contains(typeof(IUst))))
-                        {
-                            result.Add(type.Name.ToLowerInvariant(), type);
-                        }
+                        ustType = assemblyType;
```
gabriel-fallen: This will overwrite `ustType` with the last `assemblyType` satisfying the condition if there are several of those. Is it correct behaviour? Why? Or should there be an early exit?

Sources/PT.PM.Common/Reflection/ReflectionCache.cs
```diff
@@ -1,53 +1,173 @@
...
-                    .Where(prop => prop.CanWrite && prop.CanRead &&
-                                   (prop.GetCustomAttribute<JsonIgnoreAttribute>() == null || prop.Name == nameof(Ust.TextSpans)))
-                    .ToArray();
-                ustNodeProperties.TryAdd(objectType, result);
+                if (registeredTypes.Contains(baseType))
+                {
+                    return;
+                }
+
+                registeredTypes.Add(baseType);
```
gabriel-fallen: Strictly speaking this is not sufficient protection against data races in general case. For instance, one thread adds hierarchy starting with `Ust` and another starting with `Expression` before the first thread have added `Expression` to the dictionary.

If such situation could never happen with any types we need a comment explaining why this is the case.

Sources/PT.PM.Common/MessagePack/RootUstMessagePackSerializer.cs
KvanTTT: @flippaa, no, it is used only for **Ust**. Every **Dfg** is linked with **Ust**,  that's why there is no need to use this collection for **Dfg**.

KvanTTT: @gabriel-fallen , duplications are not existed, because the `HashSet` is used.

Sources/PT.PM.Common/MessagePack/RootUstMessagePackSerializer.cs
KvanTTT: Ok, I will add them.

Sources/PT.PM.Common/Reflection/ReflectionCache.cs
KvanTTT: Thanks, we should add an early exit statement. Also, I think it makes sense to add more unit tests for reflection checking.

Sources/PT.PM.Common/Reflection/ReflectionCache.cs
KvanTTT: Actually, I don't like the current implementation, because I want to get rid of situations with double addition. I will add a comment.

Sources/PT.PM.Common/Nodes/RootUst.cs
KvanTTT: It's a workaround for external jsons deserialization. I'll mark it with a comment.

Sources/PT.PM.Common/MessagePack/RootUstMessagePackSerializer.cs
```diff
@@ -20,6 +20,9 @@ public class RootUstMessagePackSerializer
         private TextSpanSerializer textSpanSerializer;
         private FileSerializer fileSerializer;
 
+        // One file can contain several RootUst nodes. For example, the file with PHP inside JavaScript inside PHP
+        // Deserialization is being occurred in depth-first manner, that's why it's possible to fill roots and parents
```
gabriel-fallen: > Deserialization is being occurred in depth-first manner

"Deserialization is (being) conducted in depth-first manner" maybe? Or "performed"?

And if we first visit parent and then descendant (`RootUst`s) that's not "depth-first". Depth-first is bottom-up and in this case we wouldn't have a reference to parent `RootUst` when we process descendant. So it's a "top-down" manner, I think.

Sources/PT.PM.Common/Reflection/ReflectionCache.cs
gabriel-fallen: Use LINQ! It makes code harder to get wrong! :smile:

Sources/PT.PM.Common/TextUtils.cs
```diff
@@ -39,22 +34,22 @@ public static int FirstIndexOf(this string str, int index, bool whitespace)
...
-            var resultTextSpan = textSpans.First();
-            if (textSpans.Count() == 1)
+            var resultTextSpan = textSpans[0];
+            if (textSpans.Count == 1)
             {
                 return resultTextSpan;
             }
 
-            foreach (TextSpan textSpan in textSpans.Skip(1))
+            for (int i = 1; i < textSpans.Count; i++)
```
gabriel-fallen: I think `foreach` version conveyed intent clearer, it's easier to miss the fact that indexing starts with 1 than `Skip(1)` call.

In theory, `foreach` version could produce slightly faster code. Anyway, if actual implementation of both interfaces is the same there would be no significant performance difference, so I don't see the point of making the interface more concrete. What was the reasoning behind the decision?

Sources/PT.PM.Common/TextUtils.cs
```diff
@@ -78,42 +73,54 @@ public static TextSpan ParseAnyTextSpan(string textSpanString, out bool isLineCo
...
-                string range = parts[0].Trim().Substring(1, parts[0].Length - 2);
-                int index = range.IndexOf("..");
+                ReadOnlySpan<char> range = location.Slice(1, location.Length - 2);
+                int index = range.IndexOf("..".AsSpan());
 
                 int start, end;
                 if (index != -1)
                 {
-                    string value = range.Remove(index);
+                    string value = range.Slice(0, index).ToString(); // TODO: It will be replaced on Span when netstandard2.1 comes
```
gabriel-fallen: Will be replaced *with* Span, you mean?

"when netstandard2.1 comes _out_" I believe...

Sources/PT.PM.Matching/MatchResultDto.cs
```diff
@@ -25,6 +25,7 @@ public MatchResultDto(MatchResult matchResult)
             TextSpan = matchResult.TextSpans.Where(textSpan =>
                 textSpan.Start >= 0 &&
                 textSpan.End <= textSpan.GetSourceFile(sourceFile).Data.Length)
+                .ToList()
                 .Union();
```
gabriel-fallen: Not the other way round? :)

If `Union` is LINQ method, I'd call it before `ToList`.

KvanTTT: No, it's not a LINQ method :)

Sources/PT.PM.Common/TextUtils.cs
KvanTTT: Rider highlighted the warning of `Potential multiple enumerations`. To get rid, the input `IEnumerable` should be transformed to list or array. That's why I decided to replace it with a more concrete interface. Moreover, `IEnumerable` produces overhead code in general.

Sources/PT.PM.Common/MessagePack/RootUstMessagePackSerializer.cs
KvanTTT: You are right: the "top-down" term is more correct.

Sources/PT.PM.Common/MessagePack/RootUstMessagePackSerializer.cs
KvanTTT: Sorry, I've read it inattentively. You are right: `uint` is missing.

PR #209, opened 2019-02-19
Sql dialect detection improvements

### LINKED ISSUE #158: Optimize automatic language detection algorithm
Take a look on the following detections libs:

* [enry](https://github.com/src-d/enry)
* [linguist](https://github.com/github/linguist)

Also, it's possible to simplify the algorithm by considering only tokens stream without parsing.

codeFather2:
fixes #158


Sources/PT.PM/LanguageDetectors/SqlDialectsLexer.g4
```diff
@@ -0,0 +1,78 @@
+lexer grammar SqlDialectsLexer;
+
+channels {ERRORCHANNEL}
+
+SINGLE_LINE_COMMENT: '--' ~('\r' | '\n')* NEWLINE_EOF   -> channel(HIDDEN);
+MULTI_LINE_COMMENT:  '/*' .*? '*/'                      -> channel(HIDDEN);
+SPACE: [ \t\r\n]+                                       -> channel(HIDDEN);
+SEICOLON: ';';
```
gabriel-fallen: Looks like `M` is missing, no?

Sources/PT.PM/LanguageDetectors/SqlDialectsLexer.g4
```diff
@@ -0,0 +1,78 @@
...
+
+MY_SQL : DELIMITER | NAMES | UNDOFILE | ENGINE | INITIAL_SIZE 
+       | REVERCE_QUOTE_ID | RETURNS;
```
gabriel-fallen: `REVERSE_QUOTE_ID`?

Sources/PT.PM/LanguageDetectors/SqlDialectAntlrParser.cs
```diff
@@ -0,0 +1,37 @@
+using System;
+using Antlr4.Runtime;
+using PT.PM.AntlrUtils;
+using PT.PM.Common;
+using PT.PM.LanguageDetectors;
+
+namespace PT.PM
+{
+    public class SqlDialectAntlrParser : AntlrParser
```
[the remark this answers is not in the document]
codeFather2: this is a good idea, because specifically in this case we are interested in the handler of case-insensitive tokens.

Sources/PT.PM/LanguageDetectors/SqlDialectsLexer.g4
```diff
@@ -0,0 +1,78 @@
...
+
+fragment DELIMITER: 'DELIMITER';
```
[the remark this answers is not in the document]
codeFather2: yes, you are)

Sources/PT.PM/PT.PM.csproj
```diff
@@ -33,4 +33,12 @@
     <ProjectReference Include="..\PT.PM.TSqlParseTreeUst\PT.PM.TSqlParseTreeUst.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Antlr4" Version="4.6.5" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Remove="LanguageDetectors\SqlDialectsLexer" />
```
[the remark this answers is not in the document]
codeFather2: I ran into a problem, which was that the file was included on the builder’s list 2 times, so there were problems compiling. All because of the fact that the file is physically located in the folder from which I hook it. It just a hack for msbuild)

Sources/PT.PM/LanguageDetectors/ParserLanguageDetector.cs
```diff
@@ -136,5 +152,32 @@ public override DetectionResult Detect(TextFile sourceFile, IEnumerable<Language
 
             return null;
         }
+        
+        private Language DetectSqlDialectDialect(SqlDialectsLexer lexer)
```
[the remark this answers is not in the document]
codeFather2: i test it on ANTLR examples and on our test files (also on some projects from github), and it's work. but I can do it by rating the rating, which tokens are more

Sources/PT.PM/LanguageDetectors/SqlDialectsLexer.g4
```diff
@@ -0,0 +1,78 @@
+lexer grammar SqlDialectsLexer;
```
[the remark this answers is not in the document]
codeFather2: If in the future we are going to support vanilla SQL, then we will need a grammar with the name SqlLexer, so that we will certainly not get confused

Sources/PT.PM/PT.PM.csproj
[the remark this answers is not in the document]
KvanTTT: Ok then.

Sources/PT.PM.AntlrUtils/AntlrBaseHandler.cs
```diff
@@ -0,0 +1,82 @@
...
+using PT.PM.Common;
+using PT.PM.Common.Files;
+
+namespace PT.PM.AntlrUtils
+{
+    public abstract class AntlrBaseHandler : ILanguageHandler
+    {
+        protected static Dictionary<Language, ATN> lexerAtns = new Dictionary<Language, ATN>();
+
+        protected static Dictionary<Language, ATN> parserAtns = new Dictionary<Language, ATN>();
```
KvanTTT: `lexerAtns` and `parserAtns` should be moved to respective Lexer and Parser classes.

Sources/PT.PM.AntlrUtils/AntlrBaseHandler.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        public static ILogger StaticLogger { get; set; } = DummyLogger.Instance;
+
+        public ILogger Logger { get; set; } = DummyLogger.Instance;
+        
+        public abstract Language Language { get; }
+        public bool UseFastParseStrategyAtFirst { get; set; } = true;
```
KvanTTT: Line break.

Sources/PT.PM.AntlrUtils/AntlrBaseHandler.cs
```diff
@@ -0,0 +1,82 @@
...
+        public static long ClearCacheFilesBytes { get; set; } = 5 * 1024 * 1024L;
+
+        public static int ClearCacheFilesCount { get; set; } = 50;
+        
+        /// <summary>
+        /// Converts \r to \r\n.
+        /// </summary>
+        /// <param name="file"></param>
+        /// <returns></returns>
+        protected virtual string PreprocessText(TextFile file)
```
KvanTTT: It's definitely a Lexer method. Move it to `AntlrLexer`

Sources/PT.PM.Common/ILanguageHandler.cs
```diff
@@ -0,0 +1,7 @@
+namespace PT.PM.Common
+{
+    public interface ILanguageHandler : ILoggable
```
KvanTTT: What is the meaning of this interface? I see it used only in `AntlrBaseHandler`. So, there is no big need.

Sources/PT.PM.AntlrUtils/AntlrDumper.cs
```diff
@@ -14,7 +14,7 @@ public override void DumpTokens(ParseTree parseTree)
         {
             var antlrParseTree = parseTree as AntlrParseTree;
 
-            IVocabulary vocabulary = ((AntlrParser)parseTree.SourceLanguage.CreateParser()).Lexer.Vocabulary;
+            IVocabulary vocabulary = ((AntlrParser)parseTree.SourceLanguage.CreateParser()).Lexer.Lexer.Vocabulary;
```
KvanTTT: Maybe introduce a new method `CreateLexer`? `Lexer.Lexer` looks weird.

Sources/PT.PM.JavaParseTreeUst/JavaAntlrParser.cs
```diff
@@ -12,26 +12,20 @@ public class JavaAntlrParser : AntlrParser
 
         public JavaAntlrParser()
         {
+            Lexer = new JavaAntlrLexer();
```
KvanTTT: It's better to handle lexers creation to the `WorkflowBase` class.

Sources/PT.PM/LanguageDetectors/ParserLanguageDetector.cs
```diff
@@ -48,6 +51,17 @@ public override DetectionResult Detect(TextFile sourceFile, IEnumerable<Language
                 return new DetectionResult(langs[0]);
             }
 
+            if (langs.Any(l => l.IsSql()))
```
KvanTTT: Now I think it's more appropriate to use `All` instead of `Any`.

Sources/PT.PM/LanguageDetectors/ParserLanguageDetector.cs
```diff
@@ -48,6 +51,19 @@ public override DetectionResult Detect(TextFile sourceFile, IEnumerable<Language
                 return new DetectionResult(langs[0]);
             }
 
+            if (sourceFile.Name.EndsWith(".sql")
+                || sourceFile.Name.EndsWith(".ddl")
+                || langs.Any(l => l.IsSql()))
+            {
+                var inputStream = new AntlrCaseInsensitiveInputStream(sourceFile.Data, CaseInsensitiveType.None);
```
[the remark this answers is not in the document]
KvanTTT: Please response. I guess we should use `CaseInsensitiveType.UPPER` here.

Sources/PT.PM/LanguageDetectors/SqlDialectsAntlrLexer.cs
```diff
@@ -0,0 +1,28 @@
...
+using PT.PM.AntlrUtils;
+using PT.PM.Common;
+using PT.PM.LanguageDetectors;
+
+namespace PT.PM
+{
+    public class SqlDialectsAntlrLexer : AntlrLexer
+    {
+        private Language detectedLanguage = Language.PlSql;
+        public override Language Language => detectedLanguage;
```
KvanTTT: Insert line break.

Sources/PT.PM/LanguageDetectors/SqlDialectsLexer.g4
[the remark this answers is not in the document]
KvanTTT: What about the first question about fragments?

Sources/PT.PM/LanguageDetectors/ParserLanguageDetector.cs
codeFather2: No, cause in some cases, when languages argument is null, we have list with all languages, and in this case if any is SQL, we need to use Any() func

Sources/PT.PM.AntlrUtils/AntlrBaseHandler.cs
codeFather2: maybe we should left only one collection in base class, with name atns?

KvanTTT: Ok, it's also possible.

Sources/PT.PM.AntlrUtils/AntlrUtils.cs
```diff
@@ -31,6 +32,16 @@ public static string GetText(this ParserRuleContext ruleContext, IList<IToken> t
             return result.ToString();
         }
 
+        public static AntlrLexer CreateAntlrLexer(this Language language)
+        {
+            var parser = language.CreateParser();
```
KvanTTT: Why do you create parser here? Use the similar `CreateLexer` here.

Sources/PT.PM/LanguageDetectors/ParserLanguageDetector.cs
```diff
@@ -136,5 +150,32 @@ public override DetectionResult Detect(TextFile sourceFile, IEnumerable<Language
...
+                switch (token.Type)
+                {
+                    case SqlDialectsLexer.T_SQL:
+                        return Language.TSql;
+                    case SqlDialectsLexer.MY_SQL:
+                        return Language.MySql;
+                    case SqlDialectsLexer.PL_SQL:
+                        return Language.PlSql;
+                    default:
+                        continue;
```
KvanTTT: There is no sense to use `continue` statement here.

Sources/PT.PM.Common/IBaseLanguageParser.cs
```diff
@@ -0,0 +1,6 @@
+namespace PT.PM.Common
+{
+    public interface IBaseLanguageParser : ILoggable
```
KvanTTT: Rename to `ILanguageParserBase` for corresponding with the general style.

Sources/PT.PM.AntlrUtils/AntlrBaseHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+using Antlr4.Runtime.Atn;
+using PT.PM.Common;
+using PT.PM.Common.Files;
+
+namespace PT.PM.AntlrUtils
+{
+    public abstract class AntlrBaseHandler
+    {
+        public AntlrMemoryErrorListener ErrorListener { get; set; }
+        public static ILogger StaticLogger { get; set; } = DummyLogger.Instance;
```
KvanTTT: Static members always should be placed before the instance ones.

KvanTTT:
Is it possible to add more tests for SQL dialect detection?

PR #210, opened 2019-02-20
Fix null ust processing

inkoit:
The function fails, for example, [here](https://github.com/PositiveTechnologies/PT.PM/blob/dev/Sources/PT.PM/WorkflowBase.cs#L254), if the result is null


KvanTTT:
Thanks for the first contribution :)

PR #211, opened 2019-02-28
CliProcessorBase Logging fixes

codeFather2:


codeFather2:
This fix will help avoid logging to the default folder, because this is not always the right way.

Sources/PT.PM.Cli.Common/CliParametersNormalizer.cs
```diff
@@ -44,7 +44,7 @@ public bool Normalize(string[] args, out string[] outArgs)
                 if (!args[argInd].StartsWith("-"))
                 {
                     error = true;
-                    Logger.LogError(new ArgumentException($"Incorrect argument", args[argInd]));
+                    NormalizationErrors.Add(new ArgumentException($"Incorrect argument", args[argInd]));
```
KvanTTT: What is the meaning of  `NormalizationErrors`? Why are they named in such a way?

KvanTTT: I got it. Since they are included in `CliParametersNormalizer`, I suggest to rename this property to `Errors`.

PR #213, opened 2019-03-04
JavaScript Esprima Parser Update

KvanTTT:


Sources/PT.PM.JavaScriptParseTreeUst.Tests/JavaScriptConverterTests.cs
```diff
@@ -11,10 +12,17 @@ public class JavaScriptConverterTests
...
+                "EnhancedRegularExpression.js",
+                "MapSetAndWeakMapWeakSet.js",
+                "Modules.js",
+                "Outdated.js"
+            };
 
             TestUtility.CheckProject(Path.Combine(TestUtility.GrammarsDirectory, "javascript", "examples"),
-                Language.JavaScript, Stage.Ust);
+                Language.JavaScript, Stage.Ust, searchPredicate: fileName =>
+                    ignoredFiles.All(ignoredFile => !fileName.Contains(ignoredFile)));
```
gabriel-fallen: Maybe `!fileName.EndsWith(ignoredFile)` will better convey the intention? And hopefully work little faster? :)

Sources/PT.PM.JavaScriptParseTreeUst/JavaScriptEsprimaErrorHandler.cs
```diff
@@ -42,27 +56,18 @@ public void ThrowError(int index, int line, int column, string message)
...
-            else
-            {
-                throw CreateException(error.Index, error.Message);
-            }
+            return result;
         }
 
         public void RecordError(ParserException error)
         {
-            Logger.LogError(CreateException(error.Index, error.Message));
```
gabriel-fallen: `RecordError` does nothing now? Is this intentional? Then why can't we remove it altogether?

KvanTTT: I removed it because of errors duplication in the log. They are handled in another method.

Sources/PT.PM.JavaScriptParseTreeUst/JavaScriptEsprimaParseTreeConverter.cs
```diff
@@ -76,14 +78,17 @@ private Ust Visit(INode node)
             {
                 if (node is Statement statement)
                     return VisitStatement(statement);
-                else if (node is Expression expression)
+
+                if (node is Expression expression)
```
gabriel-fallen: What about converting these `if`s into single `switch` statement for clarity and consistency with other converters? What do you think?

KvanTTT: Good idea.

Sources/PT.PM.JavaScriptParseTreeUst/JavaScriptEsprimaParseTreeDeclarationConverter.cs
```diff
@@ -50,10 +51,13 @@ private TypeDeclaration VisitClassDeclaration(ClassDeclaration classDeclaration)
                 ? VisitIdentifier(classDeclaration.Id)
                 : null;
 
-            var baseTypes = new List<TypeToken>();
+            var baseTypes = new System.Collections.Generic.List<TypeToken>();
```
gabriel-fallen: Don't you have `Collections` alias for that? :wink:

Sources/PT.PM.JavaScriptParseTreeUst/JavaScriptEsprimaErrorHandler.cs
gabriel-fallen: But why keep the method?

gabriel-fallen: I think you should've added a comment stating why the method is left empty...

Issue #212, opened 2019-02-28
Split IntLiteral into IntLiteral, LongLiteral, BigIntLiteral, ShortLiteral


KvanTTT:
All of them should implement `INumeric` interface. Also, `PatternIntLiteral` and `PatternIntRangeLiteral` should match them.

The following properties should be added to corresponding types:

*  `int` → `IntLiteral`
* `long` → `LongLiteral`
* `BigInteger` → `BigIntLiteral`
* `short` → `ShortLiteral`

Also, maybe it makes sense to consider unsigned literals (`ByteLiteral`, `UShortLiteral`, etc.)?

Since it's not possible to detect a literal type only basing on value (sometimes the left part with type of assignment is required), the following method can be implemented for numeric parsing:

```csharp
if (!short.TryParse(text, out short shortValue))
{
    if (!int.TryParse(text, out int intValue))
    {
        if (!long.TryParse(text, out long longValue))
        {
            if (!BigInteger.TryParse(text, out BigInteger bigInteger))
            {
                // Not a number
            }
            return new BigIntLiteral(shortValue, textSpan);
        }
        return new LongLiteral(shortValue, textSpan);
    }
    return new IntLiteral(shortValue, textSpan);
}
return new ShortLiteral(shortValue, textSpan);
```

This task is actual because there are many big numbers in sql files and we should handle them. Moreover, sometimes the difference between `long` and `int` is valuable for weakness or vulnerability detection. For instance, for implicit conversions.

KvanTTT:
Closed by #214

PR #214, opened 2019-03-05
Numeric literals refactoring

codeFather2:
These changes will help to validly handle the large numbers often found in SQL dialects


codeFather2:
#212 Fix

Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
```diff
@@ -590,15 +590,17 @@ public override Ust VisitLiteralExpression(LiteralExpressionSyntax node)
...
                     return new BooleanLiteral((bool)token.Value, node.GetTextSpan());
                 case "Int32":
-                case "UInt32":
                 case "Int16":
                 case "UInt16":
                 case "Byte":
                 case "SByte":
+                    return new IntLiteral(System.Convert.ToInt32(token.Value), node.GetTextSpan());
+                case "UInt32":
+                    return new IntLiteral((int)System.Convert.ToUInt32(token.Value), node.GetTextSpan());
```
codeFather2: Maybe we should use `LongLiteral` for `UInt32` and `BigIntLiteral` for `UInt64`. What do u think about this?

Sources/PT.PM.AntlrUtils/AntlrConverter.cs
```diff
@@ -311,13 +323,24 @@ protected Token ExtractLiteral(IToken token)
                     text = text.Substring(1, text.Length - 2);
                     result = new StringLiteral(text, textSpan);
                 }
-                else if (text.All(c => char.IsDigit(c)))
+                else if (text.All(char.IsDigit))
```
gabriel-fallen: :+1: :smiley:

Sources/PT.PM.AntlrUtils/AntlrConverter.cs
```diff
@@ -311,13 +323,24 @@ protected Token ExtractLiteral(IToken token)
...
+                    }
+                    else
+                    {
+                        result = new BigIntLiteral(BigInteger.Parse(text), textSpan);
+                    }
                 }
                 else if (text.StartsWith("0X", StringComparison.OrdinalIgnoreCase))
                 {
-                    result = new IntLiteral(System.Convert.ToInt64(text.Substring(2), 16), textSpan);
+                    result = new LongLiteral(System.Convert.ToInt64(text.Substring(2), 16), textSpan);
```
gabriel-fallen: What about BigInts in a hexadecimal representation?

Sources/PT.PM.AntlrUtils/AntlrConverter.cs
```diff
@@ -311,13 +323,24 @@ protected Token ExtractLiteral(IToken token)
...
+                    else if (long.TryParse(text, out long longValue))
+                    {
+                        result = new LongLiteral(longValue, textSpan);
+                    }
+                    else
+                    {
+                        result = new BigIntLiteral(BigInteger.Parse(text), textSpan);
+                    }
                 }
                 else if (text.StartsWith("0X", StringComparison.OrdinalIgnoreCase))
```
gabriel-fallen: What about "0x" prefix? Or `text` gets capitalized before this comparison?

Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
gabriel-fallen: `BigInt` feels a bit wasteful for `UInt64`...
Can we make `IntLiteral` and `LongLiteral` seamlessly handle both signed and unsigned integers under the hood?

Sources/PT.PM.Common/Nodes/UstVisitor.cs
```diff
@@ -178,6 +178,16 @@ public virtual T Visit(IntLiteral intLiteral)
             return VisitChildren(intLiteral);
         }
 
+        public virtual T Visit(BigIntLiteral bigIntLiteral)
+        {
+            return VisitChildren(bigIntLiteral);
```
gabriel-fallen: Sorry, what children?

Sources/PT.PM.Dsl/DslUstConverter.cs
```diff
@@ -522,19 +531,23 @@ public PatternUst VisitPatternLiteral(DslParser.PatternLiteralContext context)
...
+                }
+                else
+                {
+                    return new PatternBigIntLiteral(BigInteger.Parse(text), context.GetTextSpan());
+                }
             }
+
+            long resultValue = context.PatternOct() != null
+                ? System.Convert.ToInt64(context.PatternOct().GetText(), 8)
+                : System.Convert.ToInt64(context.PatternHex().GetText(), 16);
```
gabriel-fallen: Again, no octal or hexadecimal `BigInt`s?

Sources/PT.PM.Matching/PatternNormalizer.cs
```diff
@@ -97,19 +97,19 @@ public override PatternUst Visit(PatternIdRegexToken patternIdRegexToken)
...
+            if (patternIntRangeLiteral.MinValue == patternIntRangeLiteral.MaxValue)
             {
                 return new PatternIntLiteral(
-                    patternIntLiteral.MinValue,
-                    patternIntLiteral.TextSpan);
+                    patternIntRangeLiteral.MinValue,
+                    patternIntRangeLiteral.TextSpan);
             }
 
             return new PatternIntRangeLiteral(
```
gabriel-fallen: Why can't we just return `patternIntRangeLiteral` itself here? Why new instance which is a copy of the given one?

Sources/PT.PM.Matching/PatternVisitor.cs
```diff
@@ -119,16 +119,21 @@ public virtual T Visit(PatternIndexerExpression patternIndexerExpression)
             return VisitChildren(patternIndexerExpression);
         }
 
-        public virtual T Visit(PatternIntLiteral patternIntLiteral)
+        public virtual T Visit(PatternIntLiteral patternNumericLiteral)
         {
-            return VisitChildren(patternIntLiteral);
+            return VisitChildren(patternNumericLiteral);
```
gabriel-fallen: Does it have any children?

Sources/PT.PM.Matching/PatternVisitor.cs
```diff
@@ -119,16 +119,21 @@ public virtual T Visit(PatternIndexerExpression patternIndexerExpression)
...
         }
 
         public virtual T Visit(PatternIntRangeLiteral patternIntRangeLiteral)
         {
             return VisitChildren(patternIntRangeLiteral);
         }
 
+        public virtual T Visit(PatternBigIntLiteral patternBigIntLiteral)
+        {
+            return VisitChildren(patternBigIntLiteral);
```
gabriel-fallen: What children again? :smile:

Sources/PT.PM.AntlrUtils/AntlrConverter.cs
KvanTTT: Take attention: the `StringComparison.OrdinalIgnoreCase` option is used here. Thus there is no need to lower input string.

Sources/PT.PM.Matching/Patterns/PatternIntLiteral.cs
```diff
@@ -18,19 +18,31 @@ public PatternIntLiteral(long value, TextSpan textSpan = default)
...
         protected override MatchContext Match(Ust ust, MatchContext context)
         {
             if (ust is IntLiteral intLiteral)
             {
-                return Value == intLiteral.Value ? context.AddMatch(intLiteral) : context.Fail();
+                return Value == intLiteral.Value
+                ? context.AddMatch(ust)
+                : context.Fail();
+            }
+            if (ust is LongLiteral longLiteral)
```
gabriel-fallen: But what about `BigIntLiteral`?

Sources/PT.PM.Common/Nodes/UstVisitor.cs
codeFather2: This is a basic visitor method, by default it does nothing for any node. This class is used for inheritance if we are only interested in certain types of nodes and the rest can be skipped.

Sources/PT.PM.PlSqlParseTreeUst/PlSqlConverterVisitorHelper.cs
```diff
@@ -27,7 +28,18 @@ public override Ust VisitTerminal(ITerminalNode node)
             }
             else if (text.All(c => char.IsDigit(c)))
```
gabriel-fallen: No etha-expansion here? :smile:

Sources/PT.PM.TSqlParseTreeUst/TSqlAntlrConverter.cs
```diff
@@ -1071,7 +1072,21 @@ public Ust VisitIndex_option([NotNull] TSqlParser.Index_optionContext context)
             }
             else
             {
-                right = new IntLiteral(long.Parse(context.DECIMAL().GetText()), context.DECIMAL().Symbol.GetTextSpan());
+                string rightText = context.DECIMAL().GetText();
+                var rightTextSpan = context.DECIMAL().Symbol.GetTextSpan();
+                
+                if (int.TryParse(rightText, out int intValue))
```
gabriel-fallen: This piece of code gets repetitive and annoying - can it be extracted into separate method?

Sources/PT.PM.Common/Nodes/UstVisitor.cs
gabriel-fallen: Yep, and `VisitChildren` will return `null` when there are no children, why not just immediately return `null`?

Sources/PT.PM.AntlrUtils/AntlrConverter.cs
gabriel-fallen: Sorry, my bad! :smile:

Sources/PT.PM.Dsl/DslUstConverter.cs
```diff
@@ -522,19 +531,23 @@ public PatternUst VisitPatternLiteral(DslParser.PatternLiteralContext context)
...
-            }
-            else if (context.PatternInt() != null)
-            {
-                resultValue = long.Parse(context.PatternInt().GetText());
-            }
-            else
+            if (context.PatternInt() != null)
             {
-                resultValue = System.Convert.ToInt64(context.PatternHex().GetText(), 16);
+                var text = context.PatternInt().GetText();
```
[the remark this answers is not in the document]
codeFather2: this can be ignored, just diff is displayed not quite right

Sources/PT.PM.Matching/Patterns/PatternIntLiteral.cs
codeFather2: `PatternBigIntLiterl` handle big numbers

Sources/PT.PM.Common/Nodes/UstVisitor.cs
KvanTTT: It's default stub for `Visit` methods.

Sources/PT.PM.Matching/Patterns/PatternIntLiteral.cs
gabriel-fallen: But actual `BigIntLiteral` might be small enough to match int literal.

Sources/PT.PM.Common/Nodes/UstVisitor.cs
gabriel-fallen: Why not to change the default to just `return null`?

codeFather2: it's not possible to use `return null` because we have Generic class, but i change it to return default, where it possible

Sources/PT.PM.Matching/Patterns/PatternIntLiteral.cs
codeFather2: `BigIntLiteral` may not be small enough, since it was created specifically for processing numbers that are too large for ordinary long and int values. That is, if the number is small enough to fit in long, then it will be in `LongLiteral`

gabriel-fallen: What about languages where it's possible to explicitly specify BigInt decimals, like F#?
```fsharp
let v : bigint = 2I
```

codeFather2: my bad, did not know about such f# features

KvanTTT:
Please, rebase on dev at first.

Sources/PT.PM.Common/Nodes/UstVisitor.cs
KvanTTT: In this case, processing will be terminated for a node type with undefined `Visitor` right away. I suggest leaving it as it.

Sources/PT.PM.Matching/Patterns/PatternIntLiteral.cs
gabriel-fallen: Yeah, F# is in no way unique in this regard.

Sources/PT.PM.Common/Nodes/UstVisitor.cs
codeFather2: after painful deliberation, I decided to insist on the endurance of uniformity for all methods)

Sources/PT.PM.Common/TextUtils.cs
```diff
@@ -382,5 +384,64 @@ private static void ParseLineColumn(ReadOnlySpan<char> text, out int line, out i
...
         }
+
+        public static Literal CreateNumericLiteral(string value, TextSpan textSpan, int fromBase = 10)
+        {
+            switch (fromBase)
+            {
+                case 2:
+                    try
+                    {
+                        return new LongLiteral(Convert.ToInt64(value, 2), textSpan);
```
KvanTTT: Why don't we use `IntLiteral` for bases other than 10 (2, 8, 16)?

Sources/PT.PM.Common/UstConstantFolder.cs
```diff
@@ -144,8 +143,15 @@ private FoldResult TryFoldBinaryOperatorExpression(BinaryOperatorExpression bina
                     return result;
                 }
             }
-            else if (leftValue is long leftInt && rightValue is long rightInt)
+            else if ((leftValue is long || leftValue is int)
+                && (rightValue is long || rightValue is int))
```
KvanTTT: You forgot about `BigInt`. Moreover, I think it's better to get rid of type casts here and to split this method into 3 ones: for `int`, `long` and `BigInt` types.

Sources/PT.PM.Tests/NumberTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        
+        [Test]
+        public void Number_Recognition_Check()
+        {
+            TestUtility.CheckFile(testFileName, Stage.Ust, out RootUst ust);
+
+            Assert.IsTrue(ust.AnyDescendantOrSelf(descendant =>
+            {
+                return descendant is IntLiteral;
+            }));
```
KvanTTT: Maybe convert it to lambda expression?

```cs
Assert.IsTrue(ust.AnyDescendantOrSelf(descendant =>descendant is IntLiteral));
```

Sources/PT.PM.Common/TextUtils.cs
```diff
@@ -382,5 +384,64 @@ private static void ParseLineColumn(ReadOnlySpan<char> text, out int line, out i
...
+                        return new LongLiteral(Convert.ToInt64(value, 16), textSpan);
+                    }
+                    catch
+                    {
+                        return new BigIntLiteral(BigInteger.Parse(value, System.Globalization.NumberStyles.HexNumber));
+                    }
+                default:
+                    throw new NotSupportedException($"{fromBase} base сonversion is not supported");
+            }
+        }
```
KvanTTT: I suggest wrapping this method to `try catch` block and return default `IntLiteral` with value of `0` if exception. Conversion step should be reliable as possible.

Sources/PT.PM.Common/UstConstantFolder.cs
```diff
@@ -234,14 +240,26 @@ private FoldResult TryFoldUnaryOperatorExpression(UnaryOperatorExpression unaryO
 
             if (unaryOperatorExpression.Operator.UnaryOperator == UnaryOperator.Minus)
             {
-                if (foldResult.Value is long longValue)
+                object valueToProcess = null;
+                switch (foldResult.Value)
                 {
-                    return ProcessUnaryExpression(unaryOperatorExpression, foldResult, -longValue);
+                    case double doubleVal:
+                        valueToProcess = -doubleVal;
```
KvanTTT: Excess boxing (conversion to object). It's better to replace with return statements.

Sources/PT.PM.Common/Nodes/Tokens/Literals/INumericLiteral.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace PT.PM.Common.Nodes.Tokens.Literals
+{
+    public interface INumericLiteral
```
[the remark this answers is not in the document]
codeFather2: Ok, sure. but without `IConvertible`, cause `BigInteger` does not implement this Interface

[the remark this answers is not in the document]
KvanTTT: Ok.

Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
gabriel-fallen: We describe syntactic (unsigned) integers, not calculations. Thus `IntLiteral` class can represent both just keeping two fields - for both signed and unsigned, assigning to the right one upon construction and remembering which one to use.

More OOP way is two have two subclasses of `IntLiteral` of course...

Sources/PT.PM.Common/TextUtils.cs
gabriel-fallen: There are very few languages out there allowing arbitrary base for integer literals. And I can't recall such language among more-or-less popular ones.

PR #218, opened 2019-04-12
Python fixes

codeFather2:


Sources/PT.PM.PythonParseTreeUst/PythonAntlrConverter.cs
```diff
@@ -44,6 +44,14 @@ public Ust VisitFile_input(PythonParser.File_inputContext context)
             {
                 block.Statements.Add(Visit(child).ToStatementIfRequired());
             }
+
+            if (root.CurrentSourceFile?.Name.Length > 0)
+            {
+                var dotIndex = root.CurrentSourceFile.Name.LastIndexOf('.');
+                return new NamespaceDeclaration(new StringLiteral(root.CurrentSourceFile?.Name
```
gabriel-fallen: Could you please reformat this like
```cs
return new NamespaceDeclaration(
    new StringLiteral(root.CurrentSourceFile?.Name.Remove(dotIndex < 0 ? 0 : dotIndex)),
    block,
    context.GetTextSpan());
```
? Otherwise it's hard to follow where one argument ends and the next one starts.

Sources/PT.PM.PythonParseTreeUst/PythonAntlrConverter.cs
```diff
@@ -984,12 +992,20 @@ public Ust VisitEncoding_decl(PythonParser.Encoding_declContext context)
 
         public Ust VisitYield_expr(PythonParser.Yield_exprContext context)
         {
-            return VisitChildren(context);
+            var yielArg = context.yield_arg();
+            return yielArg == null
+                ? new YieldExpression(null, context.GetTextSpan())
+                : new YieldExpression(Visit(context.yield_arg()).ToExpressionIfRequired(), context.GetTextSpan());
```
gabriel-fallen: I think you can reuse `yielArg` here. And rename it to `yieldArg` by the way. :)

gabriel-fallen (approved):
Looks about right. :smile:

KvanTTT (approved):

PR #222, opened 2019-04-23
Implicit url in .gitmodules

inkoit:
Relative ones don't work if forked


KvanTTT:
Unfortunately, we can not accept this MR because of our internal CI system works unstable (we use mirroring for GitHub repositories) with absolute submodule paths.

Could you fork submodules too? It should work.

inkoit:
Yep, it was my prev solution, but I've marked it as the ugly one in my mind.

Closing PR

PR #223, opened 2019-04-23
Added references in TestUtils

inkoit:
Rider doesn't want to run any tests without these packages installed


KvanTTT:
That's strange because my Rider and Visual Studio recognizes tests from all test projects. `TestUtils` project does not have tests itself - it's a test foundation that is used in other `.Tests` projects.

PR #226, opened 2019-05-13
Zero allocation approaches, memory optimizations

KvanTTT:
* Use light alternatives `LightInputStream`, `LightTokenFactory`, `LightTokeSource` in ANTLR lexers.
* Mark `TextSpan`, `LineColumnTextSpan` with `readonly struct`.
* `CommentLiteral` -> `Comment`, inherit it from `Ust` instead of `StringLiteral`.
* Get rid of string allocations in `Comment` and `StringLiteral`, store only `TextSpan`.
* Get rid of storing read source files during scan, add `EventHandler<IFile> FileRead` to `WorkflowBase`.
* Decrese number of string allocations in convert and match methods.
* Replace useless array of single element `TextSpan[]` in `Ust` on single `TextSpan`.
* Remove compression code in MessagePack serialization and related tests.
* ANTLR cache: clear all ATN in time instead of language specific clearing.

Moreover: I've fixed problems with tests and now they are always passed!


Sources/PT.PM.AntlrUtils/LightInputStream.cs
```diff
@@ -0,0 +1,158 @@
...
+            }
+
+            size = input.Length;
+        }
+
+        public void Consume()
+        {
+            if (position >= size)
+                throw new InvalidOperationException("cannot consume EOF");
+            if (position >= size)
```
codeFather2: Same condition as on 60 line. Is it ok?

Sources/PT.PM.AntlrUtils/LightInputStream.cs
```diff
@@ -0,0 +1,158 @@
...
+            {
+                index = Math.Min(index, size);
+                while (position < index)
+                    Consume();
+            }
+        }
+
+        public string GetText(Interval interval)
+        {
+            int a = interval.a;
```
codeFather2: It may be better to use the names `start` and `end` instead of `a` and `num`

Sources/PT.PM.CSharpParseTreeUst/Converter/ExpressionVisitor.cs
```diff
@@ -671,7 +670,7 @@ public override Ust VisitRefTypeExpression(RefTypeExpressionSyntax node)
         {
             IdToken id = ConvertId(node.Keyword);
             Expression expr = (Expression)base.Visit(node.Expression);
```
codeFather2: maybe change to `var expr` here?

Sources/PT.PM.AntlrUtils/LightInputStream.cs
gabriel-fallen: It's obviously not OK, it's unreachable.

KvanTTT: Thank you for attentiveness :+1:

Sources/PT.PM.AntlrUtils/LightInputStream.cs
KvanTTT: Yes, it's better. The code was copied from ANTLR class, that's why it's out of style.

Sources/PT.PM.AntlrUtils/AntlrUtils.cs
```diff
@@ -16,19 +17,52 @@ namespace PT.PM.AntlrUtils
...
+        private static long processedFilesCount;
+        private static long processedBytesCount;
+        private static long checkNumber;
+        private static volatile bool excessMemory;
+
+        public static long MemoryConsumptionBytes { get; set; } = 2 * 1024 * 1024 * 1024L;
+
+        public static long ClearCacheFilesBytes { get; set; } = 5 * 1024 * 1024L;
+
+        public static int ClearCacheFilesCount { get; set; } = 50;
```
gabriel-fallen, line 26: Three static fields above are named a bit unclear. Are they limits? Or increments? I suggest that the names should reflect that.
For instance, `MaxAllocatedMemoryBytes` or `MemoryLimitInBytes`; `MaxCacheSizeInBites`; `MaxFilesInCache` or `MaxCachedFiles`.

Sources/PT.PM.AntlrUtils/LightInputStream.cs
```diff
@@ -0,0 +1,158 @@
...
+
+            CaseInsensitiveType = caseInsensitiveType;
+
+            data = new char[input.Length];
+            int i = 0;
+            while (i < input.Length)
+            {
+                if (input[i] == '\r')
+                {
+                    if (i + 1 >= input.Length || input[i + 1] != '\n')
```
gabriel-fallen: Can't this condition be merged into previous one like this `(input[i] == '\r') && (i + 1 < input.Length) && (input[i + 1] != '\n')`? Then you can get rid of duplicate else branch. And arguably the intention becomes clearer.

gabriel-fallen: LOL, I still got the logic wrong. :smile: 
OK, the translation should be literal: `(input[i] == '\r') && ((i + 1 >= input.Length) || (input[i + 1] != '\n'))`

Sources/PT.PM.AntlrUtils/LightInputStream.cs
```diff
@@ -0,0 +1,150 @@
...
+                while (Index < index)
+                    Consume();
+            }
+        }
+
+        public string GetText(Interval interval)
+        {
+            int start = interval.a;
+            int end = interval.b;
+            if (end >= Size)
```
gabriel-fallen: Is it `int end = Math.Min(interval.b, Size - 1)`? :smiley:

Sources/PT.PM.AntlrUtils/LightToken.cs
```diff
@@ -0,0 +1,74 @@
...
+        public int StartIndex { get; }
+
+        public int StopIndex { get; }
+
+        public int TokenIndex { get; }
+
+        public LightToken(LightInputStream inputStream, int type, int channel, int index, int start, int stop)
+        {
+            this.inputStream = inputStream;
+            this.type = (short) type;
```
gabriel-fallen, line 24: What's the point in accepting `int`s and truncating them to `short`s?

KvanTTT: Memory conserving. Now we have `short type` and `short channel`. Together they take 4 bytes instead of 8. It makes sense for huge projects.

Sources/PT.PM.Common/ConvertHelper.cs
```diff
@@ -0,0 +1,277 @@
...
+            this.root = root ?? throw new ArgumentNullException(nameof(root));
+            Language = root.Language;
+        }
+
+        public Token ConvertToken(ReadOnlySpan<char> span, TextSpan textSpan)
+        {
+            int spanLength = span.Length;
+            char firstChar = span[0];
+
+            if (spanLength == 1 && firstChar == '*')
```
gabriel-fallen, line 29: Looks like `&& firstChar == '*'` part is excessive here. Otherwise code in then branch makes little sense.

Sources/PT.PM.Common/Nodes/Comment.cs
```diff
@@ -0,0 +1,19 @@
+﻿using MessagePack;
+
+namespace PT.PM.Common.Nodes.Tokens.Literals
+{
+    [MessagePackObject]
+    public class Comment : Ust
```
gabriel-fallen, line 6: `Comment` is still in `Literals` namespace despite not inheriting from `UstLiteral`?

Sources/PT.PM.Matching/Patterns/PatternStringLiteral.cs
```diff
@@ -24,16 +25,31 @@ protected override MatchContext Match(Ust ust, MatchContext context)
...
+                }
+
+                return String.Equals(stringLiteral.Text)
+                    ? context.AddMatch(stringLiteral.ViewTextSpan)
+                    : context.Fail();
             }
 
             if (context.UstConstantFolder != null &&
-                context.UstConstantFolder.TryGetOrFold(ust, out FoldResult foldingResult))
+                context.UstConstantFolder.TryGetOrFold(ust, out FoldResult foldResult))
```
gabriel-fallen, line 42: Other options: `foldedResult` or even just `folded`. :smiley:

Sources/PT.PM.PythonParseTreeUst/PythonAntlrConverter.cs
```diff
@@ -806,12 +812,35 @@ public Ust VisitName(PythonParser.NameContext context)
 
         public Ust VisitNumber(PythonParser.NumberContext context)
         {
+            if (context.integer() != null)
```
gabriel-fallen, line 815: Can you please upper-case `integer()` for consistency with `FLOAT_NUMBER()` while you're editing it anyway? :wink:

gabriel-fallen:
I've missed the point when it got merged. :smile:

Sources/PT.PM.Common/Nodes/Comment.cs
KvanTTT: Actually, comments are hidden tokens, they can be completely skipped without influence compilation. In Roslyn, they are named as [syntax trivia](https://github.com/dotnet/roslyn/wiki/Roslyn-Overview#syntax-trivia) that are not inherited from any ast class. So, comments are even not `Ust`, but I left it by such way, because of some compatibility reasons.

Also, we have another meaning of literals - they can be placed to any expression.

Sources/PT.PM.PythonParseTreeUst/PythonAntlrConverter.cs
KvanTTT: It's not possible, because it comes from generated code by ANTLR. BTW, token names are started with UPPER case letter, parse tree names are started with lower case letter. Thus, `integer` means parse tree node.

KvanTTT:
I did it ASAP because of the upcoming PHDays release 😄

gabriel-fallen:
That's perfectly understandable! :smiley:

PR #227, opened 2019-05-15
FileKey in RootUst and other small fixes

KvanTTT:


Sources/PT.PM.Common/MultiMap.cs
```diff
@@ -53,6 +53,16 @@ public void Add(TKey key, TValue value)
             storageValue.Add(value);
         }
 
+        public void Add(TKey key, IEnumerable<TValue> values)
```
gabriel-fallen, line 56: Maybe rename it `AddRange` for consistency with other collections? But I'm not 100% sure about this one...

gabriel-fallen (approved):
LGTM! :+1:

Sources/PT.PM.Common/UstConstantFolder.cs
```diff
@@ -74,16 +74,15 @@ private FoldResult TryGetOrFold(Ust ust)
 
         private FoldResult TryFoldArrayCreationExpression(ArrayCreationExpression arrayCreationExpression)
         {
-            if (arrayCreationExpression.Type?.TypeText == "char" &&
-                arrayCreationExpression.Initializers.Count > 0 &&
-                (bool)arrayCreationExpression.Initializers?.All(i => i is StringLiteral))
+            if (arrayCreationExpression.Initializers?.Count > 0 &&
+                !arrayCreationExpression.Initializers.Any(i => !(i is StringLiteral)))
```
gabriel-fallen: To me, `All is StringLiteral` is clearer than `Not any not is StringLiteral`.

Sources/PT.PM.PythonParseTreeUst/PythonAntlrConverter.cs
```diff
@@ -788,19 +803,22 @@ public Ust VisitAtom(PythonParser.AtomContext context)
...
-                    };
-                }
+                var visited = Visit(context.GetChild(1));
 
-                return Visit(context.GetChild(1));
+                return new ArrayCreationExpression
+                {
+                    Initializers = visited is MultichildExpression multichild
+                        ? UstUtils.ExtractMultiChild(multichild)
+                        : new List<Expression> { visited.ToExpressionIfRequired() },
```
gabriel-fallen: To me `ToExpressionIfRequired` is a bit strange name. You require it here to be an expression, what other ifs there might be? Perhaps `ToExpressionIfNeeded` or even `AsExpression` would be a better name?

The same goes for `ToStatementIfRequired`

gabriel-fallen (approved):
OK.

Sources/PT.PM.Common/MultiMap.cs
KvanTTT: I think it's ok because of it's `MultiMap` but not `Dictionary`.

Sources/PT.PM.Common/UstConstantFolder.cs
KvanTTT: Agree. Actually Rider suggests such simplification 😄

Sources/PT.PM.Matching/MatchUtils.cs
```diff
@@ -15,14 +15,14 @@ internal static List<TextSpan> AlignTextSpans(List<TextSpan> foldedTextSpans, Li
         {
             List<TextSpan> result = new List<TextSpan>(matchesLocations.Count);
 
-            int startOffset = foldedTextSpans[0].Start;
+            int startOffset = foldedTextSpans.FirstOrDefault().Start;
```
KvanTTT, line 18: What's the point of using `FirstOrDefault` here? It's `TextSpan.Zero` here because of `TextSpan` is a struct.

Sources/PT.PM.PythonParseTreeUst/PythonAntlrConverter.cs
KvanTTT: It makes sense. I like `AsExpression` because of its small length.

PR #229, opened 2019-06-03
Recording matched UST node in MatchResult.

gabriel-fallen:
That's basically all I'm doing here. It's needed in another project.

### DIFF
--- a/Sources/PT.PM.Matching/MatchResult.cs
+++ b/Sources/PT.PM.Matching/MatchResult.cs
@@ -17,10 +17,8 @@ namespace PT.PM.Matching
 
         public TextSpan TextSpan => TextSpans.FirstOrDefault();
 
-        public MatchResult(TextFile sourceFile, PatternRoot pattern, IEnumerable<TextSpan> textSpans)
+        private MatchResult(PatternRoot pattern, IEnumerable<TextSpan> textSpans)
         {
-            Ust = null;
-            SourceFile = sourceFile ?? throw new ArgumentNullException(nameof(sourceFile));
             Pattern = pattern ?? throw new ArgumentNullException(nameof(pattern));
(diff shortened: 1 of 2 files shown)


Sources/PT.PM.Matching/MatchResult.cs
```diff
@@ -27,6 +31,20 @@ public MatchResult(TextFile sourceFile, PatternRoot pattern, IEnumerable<TextSpa
...
+            SourceFile = Ust.CurrentSourceFile;
+            Pattern = pattern ?? throw new ArgumentNullException(nameof(pattern));
+            TextSpans = textSpans?.ToArray() ?? throw new ArgumentNullException(nameof(textSpans));
+
+            TextSpan lastTextSpan = textSpans.LastOrDefault();
+            if (lastTextSpan != default)
+            {
+                Suppressed = MatchUtils.IsSuppressed(SourceFile, lastTextSpan.GetLineColumnTextSpan(SourceFile));
+            }
+        }
```
KvanTTT, line 42: Can the common code be extracted to the protected constructor? I think so.

Sources/PT.PM.Matching/PatternRoot.cs
```diff
@@ -149,7 +149,7 @@ private static void TraverseChildren(PatternUst patternUst, Ust ust, MatchContex
                 }
                 else
                 {
-                    var match = new MatchResult(ust.CurrentSourceFile, context.PatternUst, context.Locations);
+                    var match = new MatchResult(ust, context.PatternUst, context.Locations);
```
codeFather2, line 152: And what if the positions of the text spans do not correspond to the current `ust`, for example, if we have `PatternAny`, where the matched node may be the `root`.

gabriel-fallen: How would I know? As you see, it was programmed like that long before my change. :)

PR #231, opened 2019-06-03
Conversion during parsing without building parse tree

### LINKED ISSUE #228: Implement conversion during parsing without ParseTree generation
We should turn off the option `BuildParseTrees = false` in ANTLR parser and build our UST by listeners without ANTLR parse tree at all. This allows conserving memory and improving performance. To reach this, we should write listener that build tree in ascending direction, from tokens to roots.

KvanTTT:
* Implemented `AntlrListenerConverter` listener that contains base logic of UST constructing. Specifically, `EnterEveryRule`, `ExitEveryRule`, `VisitTerminal`, `VisitErrorNode`. Implemented AntlrParserConverter that inherits `AntlrParser`.
* Implemented POC for `PL/SQL` language, see `PlSqlAntlrParserConverter`. Removed outdated `PlSqlAntlrConverter`.
* Implemented universal ANTLR token types converter. Now ANTLR tokens are defined in a declarative way, see abstract properties in `AntlrParser`.
* `ConvertToken` now can also return `Keyword`, `Operator`, `Punctuator`.

fixes #228


KvanTTT:
Do not merge. Unfortunately, I forgot to create it as a draft PR.

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
```diff
@@ -0,0 +1,255 @@
...
+
+namespace PT.PM.AntlrUtils
+{
+    /// <summary>
+    /// Generates UST node at each `ExitEveryRule` method
+    /// If needed, use concrete Listener with overriden methods for certain contexts
+    /// </summary>
+    public abstract class AntlrListenerConverter : IParseTreeListener, ILoggable
+    {
+        private static readonly char[] trimChar = {'\''};
```
gabriel-fallen, line 17: `trimChars`? Otherwise why using an array?

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
```diff
@@ -0,0 +1,255 @@
...
+
+        private int childNodesIndex = -1;
```
gabriel-fallen, line 19: `childrenNodesIndex` strictly speaking... :)

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
```diff
@@ -0,0 +1,255 @@
...
+        private readonly List<Usts> childNodes = new List<Usts>();
```
gabriel-fallen, line 20: The same here. :)

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
```diff
@@ -0,0 +1,255 @@
...
+        private readonly AntlrParserConverter antlrParserConverter;
+        private readonly AntlrConvertHelper convertHelper;
+        protected readonly RootUst root;
+
+        public ILogger Logger { get; set; } = DummyLogger.Instance;
+
+        public bool IsParseTreeExisting { get; set; }
```
gabriel-fallen, line 27: `IsParseTreeExists` maybe? I'd even drop the `Is` prefix.

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
```diff
@@ -0,0 +1,255 @@
...
+        protected AntlrListenerConverter(TextFile sourceFile, AntlrParserConverter antlrParser)
+        {
+            root = new RootUst(sourceFile, antlrParser.Language);
+            antlrParserConverter = antlrParser;
+            convertHelper = new AntlrConvertHelper(root, antlrParser);
+            PushNew(null);
+        }
+
+        /// <summary>
+        /// Calls at the end of conversion process.
```
gabriel-fallen, line 38: "Gets called at the end of conversion process" I suppose?

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
```diff
@@ -0,0 +1,255 @@
...
+        /// </summary>
+        public RootUst Complete()
+        {
+            ConvertChildren();
+            root.Node = GetChildren()[0];
+            PopChildren();
+            Clear();
+            root.FillAscendants(); // TODO: should be filled during conversion
```
gabriel-fallen, line 46: Are you sure it's "ascendants" and not "descendants"?

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
```diff
@@ -0,0 +1,255 @@
...
+            for (int i = 0; i < children.Count; i++)
+            {
+                if (children[i] is T ust)
+                {
+                    child = ust;
+                    return i;
+                }
+            }
+
+            child = default;
```
gabriel-fallen, line 94: What about `child = null` for clarity?

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
```diff
@@ -0,0 +1,255 @@
...
+            where T : class
+        {
+            List<Ust> children = GetChildren();
+
+            if (index < 0 || index >= children.Count)
+            {
+                return false;
+            }
+
+            if (!(children[index] is T ust))
```
gabriel-fallen, line 108: That's a little bit confusing naming as long as `T` doesn't have a `Ust` constraint...

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
```diff
@@ -0,0 +1,255 @@
...
+                if (result)
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        protected bool CheckChild<T>(int tokenType, int index)
```
gabriel-fallen, line 131: Shouldn't `bool CheckChild<T>(int[] tokenTypes, int index)` call this method instead of inlining the same code?

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
```diff
@@ -0,0 +1,255 @@
...
+            where T : class
+        {
+            List<Ust> children = GetChildren();
+
+            if (index < 0 || index >= children.Count)
+            {
+                return false;
+            }
+
+            if (!(children[index] is T ust))
```
gabriel-fallen, line 141: That slightly misleading naming again...

Sources/PT.PM.AntlrUtils/AntlrParser.cs
```diff
@@ -6,119 +6,196 @@
...
+        public abstract int[] CommentTokenTypes { get; }
+
+        protected abstract string ParserSerializedATN { get; }
 
         protected abstract Parser InitParser(ITokenStream inputStream);
 
         protected abstract ParserRuleContext Parse(Parser parser);
 
-        protected abstract AntlrParseTree Create(ParserRuleContext syntaxTree);
+        protected abstract AntlrParseTree CreateParseTree(ParserRuleContext syntaxTree);
```
gabriel-fallen, line 59: I think that's a misleading name of a parameter now.

Sources/PT.PM.Common/LanguageUtils.cs
```diff
@@ -114,6 +115,9 @@ static LanguageUtils()
 
         public static bool IsAntlr(this Language language) => LanguageInfos[language].HasAntlrParser;
 
+        public static bool IsParserConverter(this Language language) =>
```
gabriel-fallen, line 118: `HasParserConverter` maybe?

gabriel-fallen (changes requested):
I think it would be beneficial if you addressed my readability concerns, though it's not strictly necessary.
And I've read only first half of the changes. :smiley:

gabriel-fallen:
I missed the point when you've merged the PR. Again. :smile:

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
KvanTTT: Because of method accepts an array, not a single value.

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
KvanTTT: Fair remark. Actually, this method fills `Parent` and `Root` properties for all descendants and self. Maybe rename to `FillParentAndRootForDescendatnsAndSelf`?

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
KvanTTT: `T` doesn't have `Ust` constraint because of T can be interface such as `IOperatorOrPunctuator`. And unfortunately, C# does not have discriminated unions :) But actually `T` is always `Ust` or its successor.

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
KvanTTT: I've split these methods because of array allocation every time. It's not very optimal but not critical.

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
gabriel-fallen: I see. Then I guess renaming to plural form makes even more sense.

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
gabriel-fallen: I'd say `FillParentAndRootForSelfAndDescendatns`. :smiley:

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
gabriel-fallen: Then I suggest adding a comment saying the same you're saying here "for future generations to know this sacred knowledge". :smile:

Sources/PT.PM.AntlrUtils/AntlrListenerConverter.cs
KvanTTT: I prefer my variant because of Roslyn team use the same order: https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.operations.operationextensions.descendantsandself?view=roslyn-dotnet

PR #235, opened 2019-07-04
Got rid of drawback with TextSpan MsgPack deserialization

flippaa:
Since key has been chosen as unique ID, it's redundant to support multiTextSpan case(currently only with MessagePack format)

### DIFF
--- a/Sources/PT.PM.Common/MessagePack/RootUstMessagePackSerializer.cs
+++ b/Sources/PT.PM.Common/MessagePack/RootUstMessagePackSerializer.cs
@@ -157,9 +157,8 @@ namespace PT.PM.Common.MessagePack
             foreach (PropertyInfo property in serializableProperties)
             {
                 object serializeObject = property.GetValue(value);
-                Type propertyType = property.PropertyType;
 
-                newOffset += SerializeObject(ref bytes, newOffset, propertyType, serializeObject);
+                newOffset += SerializeObject(ref bytes, newOffset, property.PropertyType, serializeObject);
             }
 
             return newOffset - offset;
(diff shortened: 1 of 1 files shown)


Sources/PT.PM.Common/MessagePack/RootUstMessagePackSerializer.cs
```diff
@@ -390,6 +370,13 @@ private object DeserializeObject(byte[] bytes, int offset, Type type, out int re
                 if (MessagePackBinary.IsNil(bytes, offset))
                 {
                     MessagePackBinary.ReadNil(bytes, offset, out readSize);
+                    if (type is TextSpan)
+                    {
+                        var test = type.GetDefaultValue();
+                        var ts = (TextSpan)test;
+                        var isZero = ts == TextSpan.Zero;
+                        var a = 1;
```
KvanTTT: What is it?

Sources/PT.PM.Common/MessagePack/RootUstMessagePackSerializer.cs
```diff
@@ -157,18 +157,7 @@ private int SerializeUst(ref byte[] bytes, int offset, Ust value)
...
-                if (property.Name == nameof(Ust.TextSpan))
-                {
-                    serializeObject = new[] {(TextSpan) serializeObject};
-                    propertyType = typeof(TextSpan[]);
-                }
-                else
-                {
-                    propertyType = property.PropertyType;
-                }
+                Type propertyType = property.PropertyType;
```
KvanTTT: This var can be removed (paste `property.PropertyType` directly to method argument)

